PR #7, opened 2023-12-28
Add "Drop" interaction prompt when carrying an object

pcbeard:
This provides a way to drop a carried item.

I noticed that there was no obvious way to drop the box once it was picked up. There was no support for setting a prompt while carrying an object. This provides a simple mechanism for that, by always sending the "interaction_prompt" signal while an object is being carried.


pcbeard:
My latest commit simply displays a "Drop" prompt whenever an object is being carried.

Phazorknight (approved):

PR #10, opened 2023-12-29
Remove .godot directory, update .gitignore

pcbeard:
Godot 4 recommends not tracking the .godot directory in its auto-generated .gitignore directory.


pcbeard:
This is needed because each time I open the Cogito project and change which resource is currently selected, many files were being dirtied in the .godot directory.

Phazorknight (approved):

PR #14, opened 2024-01-06
Cache all component paths from the code into @onready variables

pcbeard:
Each use of a $ component path is a call to get_node(), so caching these up front should speed things up.


Phazorknight (approved):
Thank you!
Will bring this into main, though I'll probably update it a bit to fit GDscript naming conventions (mostly using snake case instead of camel case).

PR #15, opened 2024-01-06
Refactor test motion

pcbeard:
Cache allocations of params and test_motion_result

Instead of allocating these on each call to _physics_process(),
allocate once using @onready variables.

Also refactor calls to PhysicsServer3D.body_test_motion() to call
a helper function, test_motion() to make the code smaller.


pcbeard:
Here's a performance trace taken with these changes:

refactor_test_motion

Phazorknight (approved):
Looks good, thanks for these tweaks.

Issue #16, opened 2024-01-07
Bi-directional door animations


pcbeard:
I've been working with these [house assets](https://loafbrr.itch.io/houses-and-props) and realized that the default animations it uses which open a door when a player character is nearby are a bit awkward to use when the doors open toward the character. Toward that end, I've prototyped a solution that I would love to see incorporated into Cogito, if possible. You can see what it looks like in [this mastodon post](https://sfba.social/@pcbeard/111712548072890686).

The main idea would be to have the door swing inward if approached from one side, and outward from the other. I implemented this by adding another Area3D to determine which side the of the door the player is approaching from, and managing a simple state machine to run the animations.

pcbeard:
Here's the GDScript code I wrote to implement this logic:

```GDScript
extends Node3D

@onready var animation = $AnimationPlayer

enum { CLOSED, CLOSING, OPENING, OPEN }
var door_state = CLOSED
var in_outside_area = false
var in_inside_area = false
var closing_animation = ""

func check_door_state():
	match door_state:
		CLOSED:
			if in_outside_area:
				door_state = OPENING
				animation.play("open_inward")
				closing_animation = "open_inward"
			elif in_inside_area:
				door_state = OPENING
				animation.play("open_outward")
				closing_animation = "open_outward"
		OPENING:
			pass
		CLOSING:
			pass
		OPEN:
			if !in_outside_area and !in_inside_area:
				door_state = CLOSING
				animation.play_backwards(closing_animation)

func _on_animation_player_animation_finished(anim_name):
	match door_state:
		OPENING:
			door_state = OPEN
		CLOSING:
			door_state = CLOSED
	check_door_state()

func _on_outside_area_3d_body_entered(body):
	if body is CharacterBody3D:
		in_outside_area = true
		check_door_state()

func _on_outside_area_3d_body_exited(body):
	if body is CharacterBody3D:
		in_outside_area = false
		check_door_state()

func _on_inside_area_3d_body_entered(body):
	if body is CharacterBody3D:
		in_inside_area = true
		check_door_state()
		
func _on_inside_area_3d_body_exited(body):
	if body is CharacterBody3D:
		in_inside_area = false
		check_door_state()
```

Phazorknight:
Thanks for this.
Your script is based on how the doors work within that asset pack and the included script. As this is a different approach on how COGITO is set up, I'm not going to include it in the repo.

That being said: I understand the request was able to add this functionality to COGITO in a different way.
[...]

PR #22, opened 2024-01-16
Add option for bi-directional door swing (using rotation tween)

grgp:
This adds the ability for doors that rotate using the Tween Parameters to open bi-directionally, depending on the position of the interactor (player). It also adds an example door in the demo scene, as you can see in the video.

Note that this PR's implementation is based on the existing `open_rotation_deg` and `closed_rotation_deg`, which seems to denote the global transform's rotation, not the local transform's.

image

https://github.com/Phazorknight/Cogito/assets/12206156/0413e050-ea0b-455c-aa46-44ef970bf8fc

Also for reference, there was an issue [here](https://github.com/Phazorknight/Cogito/issues/16) for bi-directional door animations. This PR is different because it doesn't use custom animations.


Phazorknight:
Thank you! this looks super elegant codewise.

The door script has had some changes in the experimental branch, so I will try to properly review this today or tomorrow and then hopefully merge it in.

Phazorknight (approved):

PR #50, opened 2024-02-11
Sprint, crouch, and walk footstep volume

ac-arcana:
Volume of footsteps is adjusted depending on if player is walking, sprinting, or crouching. There are variables in the player script to adjust them - walk_volume_db, sprint_volume_db, and crouch_volume_db


Phazorknight (approved):
Looks good!

PR #58, opened 2024-02-12
Dynamic Footstep System

ac-arcana:
Here is my Dynamic Footstep System. It works well in my tests. There is a demo scene included. All of the resources, assets, and scripts are contained in the DynamicFootstepSystem folder.
I did make a small change to the player script to support the system, but the system is optional and if it is not specified the player script will work normally and play generic footstep sounds.


Phazorknight:
This works great, that's exactly what I imagined.

Only a small note, then I think we can merge this into _main_:
- The added player.tscn is confusing. As your system already has a fallback sound, I see no reason keeping the old one around. Would you mind applying the new FootstepPlayer and it's script to the main player.tscn ? Feel free to include the needed changes from the player.gd script, I took a look and they seem very minor.

- In your test scene the StaminaComponent or at least the StaminaHUD doesn't seem to work? It still works in the COGITO_01_Demo.tscn so maybe some hookup got disconnected somehow.

ac-arcana:
Yes, I can get rid of the added player scene and modify the main player scene. I didn't want to change your base player without express approval.
I'm not sure why the stamina component isn't working, I'll try to figure that out. I did the basic structure of including the player, player_HUD, and pause menu and linking them together.... I'll post once I've got some more commits for this.

ac-arcana:
Def don't merge this yet, bringing these changes into the demo scene there is a bug I'll need to fix quickly

ac-arcana:
I didn't properly support static meshes and rigidbodies. Also didn't support all of the CSG shapes. I'm working on these now.

ac-arcana:
Okay I'm looking at adding support for the box in the demo right now, with the wood sounds. It looks like the box materials are inside of the box scene. I'll need to save the materials outside of the box in order to set the footstep effects. Is that alright?

Phazorknight (approved):
Amazing work!

PR #70, opened 2024-02-16
Add multiple surface functionality for meshes in Footstep System

### LINKED ISSUE #65: New surface material footstep system crashes on collisions with ArrayMesh
I use qodot for my level building which uses an ArrayMesh for every level build. The second I try to move it causes a crash with the new system. I noticed it works fine on the demo scene so I'm assuming based on the information that it is only looking for normal MeshInstances.
image

FailSpy:
Pinging @ac-arcana here for a look, as well as @Bezoika

This is my attempt at implementing handling multiple surfaces. I don't currently have a way implemented to handle multiple MeshInstance3D children, so just go for the first one.

However, with a MeshInstance3D that has an ArrayMesh or multiple surfaces, it should work. I've tested this with as many cases I can think of, but I'm doubtless missing some. 

This will try very hard to validate that there's more than one surface material needing to be thought about, and if so:
1. Find the face the player is standing on
2. If a face is found, note the vertices for that face, and translate the Vector3s into integers on a separate grid system
3. Then cycle through the vertices of each "surface" of the mesh. If it matches the grid coordinates of the found face, then it marks that surface as the correct one and returns its material.

This is far from perfect in implementation, but it's surprisingly performant on my machine, curious to know how it works for others. I did a similar stress test to that video Bezoika posted of a heavily subdivided plane with two surfaces, and got an occasional stutter but really didn't feel much other than that.

There's a whole bunch of ways I could imagine optimizing it, but again, want to know how it works for others before spending the time optimizing.

This would fix #65


FailSpy:
Oh, worth noting: I also added a check for CSGCombiner3D, as that would cause a crash similarly to ArrayMeshes.

ac-arcana:
When I tested with CSG the combiner seemed to just come back as a CSG mesh, but I'm sure a check isn't going to hurt anything and there might be a way to get it to return a combiner that I missed.

I won't have much time to look at this for the next week or so, but feel free to out a pull request in for your improvements for Phasorknight to review

ac-arcana:
I finally had time to at least read the code, looks all good to me. I would like to test it with Qodot when I get a chance as both Bezoika and myself us it.
@Bezoika have you had a chance to see if this works with Qodot?

Phazorknight (approved):
Trusting AC-Arcana's judgement and approving this.

FailSpy:
> I finally had time to at least read the code, looks all good to me. I would like to test it with Qodot when I get a chance as both Bezoika and myself us it. @Bezoika have you had a chance to see if this works with Qodot?

Thanks for reviewing it. If issues do arise/are discovered with this system in particular I'm happy to jump in on those.

Issue #79, opened 2024-02-22
Crouch jumping broken, or intentional ?


ksjfhor:
Hi, 
when you have crouch_jumping active (like the standard when opening up the project).
Do the crouch jumping but let go of the crouch_button ("c"), shortly after jumping ("space") you gain height to nearly the point where you would regulary land when you jump.

Is this intended ?
For me it looks like the JumpCoolDownTimer is not working.

https://github.com/Phazorknight/Cogito/assets/87974401/a589c0b7-ec9b-4f90-9ff4-170da9eb989a

At first I just crouched and jumped for you to see a marking with the target, otherwise it would be hard to see / recognize if not played by yourself.

Phazorknight:
Will take a look myself but from your video this looks to me as working as intended.
Crouch-jumping aka duck-jumping has been a staple in quite a few games, and I think the first person controller was explicitly made to support it.

I think the jump cooldown timer is only used to avoid retriggering jumps on bad collisions.

FailSpy:
@Phazorknight I have a branch I'm working on to address the visual jitter as well as some improvements for slide jumping

FailSpy:
I would say this is very much intentional. I tried to make jumping consistent for crouch jumps and regular jumps by saying that you wind up with the players feet off the ground by exactly the same amount

ksjfhor:
Ty both. I was not sure as it felt not right for me, I am looking forward for the [PR#82 ](https://github.com/Phazorknight/Cogito/pull/82) :)

Phazorknight:
Considering this solved with the merge from #82 / https://github.com/Phazorknight/Cogito/commit/001c48f0dbab6cb4a2450c5d8e988fdb341c893b.

PR #82, opened 2024-02-22
Improve in-air behavior for crouch and slide jumps

FailSpy:
This is a little prescriptive to what I found "feels right", but in response to #79, I've modified how crouch and slide jumps feel and look.

In #79, @ksjfhor found that un-crouching mid-air would look as though there's a sudden upwards velocity to exactly where you'd expect the jump to be if you had done it standing. It is merely a visual effect from the camera moving from the crouched height to the standing height, the jump is functionally exactly the same. The only thing that changes is the camera height, and the collision to make collisions consistent with what the player is seeing.

I personally find this behavior makes for consistent functionality both as a player and as a dev trying to find the right game feel. When crouch jumping, the characters *feet* comes off the ground by exactly the same amount whether crouched or standing. 

To address the visual jitter, I made it so you can't change your standing/crouching state visually whilst jumping. This feels a little prescriptivist, but I think is the neatest solution for right now with minimal compromise in terms of actual game functionality and feel. If you're making a parkour game however, you may find you need to alter this behavior with a custom solution to resolve this, but that likely would've been the case anyways.

I also changed the in-air behavior. If you're crouched and in-air, you will consistently have the crouch movement speed, unless you're jumping from a slide. If you're slide jumping, your character will largely be registered as being in the "standing" state, which also includes having the camera in the standing state. 

I've added 2 parameters for devs to be able to dial in their game feel:
`CROUCH_JUMP_VELOCITY` - what force your jump should have if crouch jumping
`SLIDE_JUMP_MOD` - modifier to how strong your jump is coming from a slide

### DIFF
--- a/COGITO/Scripts/player.gd
+++ b/COGITO/Scripts/player.gd
@@ -440,8 +440,12 @@ func _physics_process(delta):
 	else:
 		head.position.y = lerp(head.position.y, 0.0, delta * LERP_SPEED)
 		carryable_position.position.y = lerp(carryable_position.position.y, initial_carryable_height, delta * LERP_SPEED)
-		standing_collision_shape.disabled = false
-		crouching_collision_shape.disabled = true
+		if head.position.y < CROUCHING_DEPTH/4:
+			crouching_collision_shape.disabled = false
+			standing_collision_shape.disabled = true
+		else:
+			standing_collision_shape.disabled = false
+			crouching_collision_shape.disabled = true
 		sliding_timer.stop()
 		# Prevent sprinting if player is out of stamina.
 		if Input.is_action_pressed("sprint") and is_using_stamina and stamina_component.current_stamina > 0:
--- a/COGITO/Scripts/player.gd
+++ b/COGITO/Scripts/player.gd
@@ -69,6 +69,7 @@ var is_showing_ui : bool
 @export var AIR_LERP_SPEED = 6.0
 @export var FREE_LOOK_TILT_AMOUNT = 5.0
 @export var SLIDING_SPEED = 5.0
+@export var SLIDE_JUMP_MOD = 1.5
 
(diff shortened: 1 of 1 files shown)


Phazorknight:
I saw they were still recent commits, so I held off. Let me know when I can review.

FailSpy:
You're all set to review actually, just wrapped it up. The more recent commits were minor addendums to the overall PR.

Phazorknight:
Gave this a test and noticed that when I sprint > crouch > jump, I leap higher than with a regular jump (see video).
I'd qualify this as unexpected behavior.

https://github.com/Phazorknight/Cogito/assets/70389309/53acb499-f35a-435e-8c9f-27c911f90ac3

FailSpy:
So, how would you like that to be handled? 

That's the slide jump modifier, which I've not changed the factor of, just split it out into its own variable `SLIDE_JUMP_MOD`. If you change that to 1.0, you'll get normal jump height on doing this.

The reason it's higher than without this branch is I've made it so slide jumps put you into standing mode. I can make it so you stay at crouched height instead visually, though ultimately you're still coming off the ground higher than a regular jump.

COGITO/Scripts/player.gd
```diff
@@ -537,10 +559,12 @@ func _physics_process(delta):
 			animationPlayer.play("jump")
 			Audio.play_sound(jump_sound)
 			if !sliding_timer.is_stopped():
-				velocity.y = JUMP_VELOCITY * 1.5
```
FailSpy, line 540: @Phazorknight this is where you can see the original jump being made higher when jumping from a slide.

Phazorknight:
Oh, apologies, I didn't realize this was an intended modifier. That's what I get for rushing.
Changing the modifier did what you said and I was able to get the expected behaviour.

Will finish my review after lunch.

Phazorknight (approved):
Looks good! Really nice to have these new exported properties.
Thank you!

PR #85, opened 2024-02-23
QOL: Add keypad shortcuts so users can type with keyboard

FailSpy:
Added shortcuts to the keypad buttons so now when users type numbers on their keyboard or numpad, it will press the corresponding UI button.

Enter serves as 'E' on the keypad
Backspace or Joypad action right button serves as 'C' on the keypad

### DIFF
--- a/COGITO/InventoryPD/UiScenes/hot_bar_inventory.gd
+++ b/COGITO/InventoryPD/UiScenes/hot_bar_inventory.gd
@@ -6,7 +6,7 @@ const Slot = preload("res://COGITO/InventoryPD/UiScenes/Slot.tscn")
 @onready var h_box_container = $MarginContainer/VBoxContainer/TopRow
 
 
-func _input(event):
+func _unhandled_input(event):
 	# Handles Gamepad Hotbar input
 	if not visible:
 		return
--- a/COGITO/PrefabScenes/keypad_prefab.tscn
+++ b/COGITO/PrefabScenes/keypad_prefab.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=15 format=3 uid="uid://i4rqcx7lyrfd"]
+[gd_scene load_steps=50 format=3 uid="uid://i4rqcx7lyrfd"]
 
 [ext_resource type="Script" path="res://COGITO/Scripts/Cogito_Keypad.gd" id="1_iwk1o"]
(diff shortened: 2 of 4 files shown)


Phazorknight:
Thanks for this.
Unfortunately, these inputs overlap with default input_map_actions, causing the player to trigger these actions in addition to the keypad operations.

Examples:
- Pressing 1,2,3,4 are default shortcuts for using the items in the inventory hotbar.
- Pressing E overlaps with interact2. If the keypad is set to be opened with interact2 it closes it.

FailSpy:
Just to clarify, 'E' is in reference to the 'E' button *on the Keypad UI*. There is no usage of the 'E' keyboard button in this change. I'm using keyboard 'Enter' to press the keypad UI's 'E' button, which it doesn't look like there's any inputs currently using.

But very good point on those default number shortcuts. 
Considering that currently the keypad UI captures and prevents inputs such as moving and firing your gun, would it be okay for me to add a change that does similarly for the number shortcuts? With respect to Godot's default input consumption methods.

FailSpy:
I've just added a change that keeps track of focus and consumes the input if the keypad is visible and in focus.

Phazorknight:
>Just to clarify, 'E' is in reference to the 'E' button on the Keypad UI. There is no usage of the 'E' keyboard button in this change. I'm using keyboard 'Enter' to press the keypad UI's 'E' button

Oops, yeah you're right. I kinda confused myself.

I did notice tho that when navigating the number buttons with the keyboard arrow keys and then "pressing" those buttons with the enter key, it adds the number to the keypad (as expected) but also triggers the keypad UI 'E' button (as inteded by your change). I think this is a pretty minor thing and potentially easily fixable, but wanted to mention it.

Everything else works great otherwise!

FailSpy:
@Phazorknight Give it a shot now. Provided I've done things right:
If you've been using your keyboard to type the code, then Enter will work to press the 'E' key on the keypad and nothing else. 
If you've been manually navigating the buttons and pressing Enter on each, it will not trigger the 'E' key on the keypad.

It's also now using Godot's default 'ui_accept' action input mapping.

CogitoButton nodes now have an optional switch to 'accept contextual enter' which will provide this functionality to any UI panel that might want it, and the "context" can be set by using Control's shortcut_context

Phazorknight (approved):
Thank you! Amazing work!

PR #88, opened 2024-02-28
Lightzone Switch component

kk1201:
I noticed that the lightzone script didn't have any functionality to respond to light switch events. I added a `LightzoneSwitchComponent.gd` script to solve this. Parenting it to a `LightzoneComponent` and connecting the `switched` signal from a `Cogito_Switch.gd` node to the `_on_switched` function, this should should make the parent lightzone automatically follow the switch's state by enabling and disabling the `Area3D::monitoring` property.

**Reason**
Small change, follows composition, I think it makes sense for it to be included in the template since it's a big feature of a lot of immersive sims, allows for new gameplay and level design options :)

**No side effects**
Will only affect lightzones, and not connecting the signal or setting the switchable flag to false will make the lightzone behave as previously. `CogitoSwitch.gd::is_on` initial state is respected. I added a `LightzoneSwitchComponent` to the `LightzoneComponent` prefab and added a `LightzoneComponent` child to the `CeilingLamp` prefab as an example but feel free to make me restore them to their previous state.

**Downsides**
Uses `get_parent()`, which violates pure composition. However the use-case is incredibly obvious based on the name and documentation, and it shouldn't pollute anything else, so I feel like using it in this limited scope makes sense.

---
https://github.com/Phazorknight/Cogito/assets/149746050/a2699cb9-14d0-4613-9b53-e7327fd567bd


Phazorknight (approved):
This is awesome.
I have neglected looking into Lightzone's working on other stuff.
Having the Lightzones respond to switch state makes perfect sense and your implementation is great.
Approved!

Issue #81, opened 2024-02-22
Scalable UI for full screen


cmr624:
Is it possible to make the UI elements scaleable, so it remains its same size as in the default viewport (not full screen), when scaled up? the options menu, and hotbar are pretty unusable at full screen. 

I'm new to Godot and not sure how to achieve this with the UI system - or if an issue here is an appropriate place to ask about this - but wondering what should the approach be for this project? Thanks!!

cmr624:
Would love to help or learn as you work, when you are working on this. let me know if there are any resources you find?

grgp:
Hey I'm trying to implement this feature here, please take a look!
https://github.com/Phazorknight/Cogito/pull/90

Phazorknight:
Basic UI scaling is implemented, so I'm going to close this.

PR #90, opened 2024-03-01
[WIP] Add GUI scale setting to options

grgp:
Addresses https://github.com/Phazorknight/Cogito/issues/81

Added slider here:
godot-screenshot-gui

Changed project settings:
image

I used the official examples repo [here](https://github.com/godotengine/godot-demo-projects/blob/master/gui/multiple_resolutions/main.gd) as a reference. Still marking it as a work in progress because I haven't handled the case if the player simply clicks—not drag—the slider button.

It works right now, but it's kind of wonky.


grgp:
Ah nice findings! Huh this is where my knowledge runs out tbh. I would assume since the 2D elements are from built-in Godot (vector and text, instead of image textures), it would scale perfectly no matter what scale we give it.

I'll leave this PR open for now in case anyone else has ideas. I'll ask on the Godot Discord, actually.

PR #92, opened 2024-03-02
Remove .godot folder and its remaining files still in git tracking

grgp:
### Pull Request

Despite `.godot` being put in .gitignore, [some files](https://github.com/Phazorknight/Cogito/tree/main/.godot) are still tracked by git. I think because the line `.godot` was only added to the .gitignore file sometime after the project was initialized. Git wouldn't track new files added to the `.godot` folder, but it still tracks the old files, which can cause some issue.

To remove the files, what I did was to run this.
```
git rm -r --cached --ignore-unmatch .godot
```

After this PR is merged, most users would need to reimport a few files, but it should work fine. Worse case scenario, they'll need to delete their `.godot` folder.

---

### Background issue

A lot of times after I sync my fork to the latest commit, I'll get these changes after I open the project in Godot.

```
On branch main
Your branch is up to date with 'origin/main'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   COGITO/DemoScenes/COGITO_04_Demo_Lobby.tscn
        modified:   COGITO/DemoScenes/COGITO_05_Demo_Laboratory.tscn
        modified:   COGITO/Theme/Cogito_Theme_A.tres
```

This is despite the fact I didn't do anything to the scene. I assumed it has to do something with the `.godot` folder. But since some files are still tracked by git, I couldn't just remove the `.godot` folder because I'll see something like this.

image

This problem should be fixed after this PR is merged, though.


Phazorknight (approved):
Thanks!
I think in one commit a while back some .godot files snuck back in and I was just stuck with them ever since. 😅 
Just gonna approve this.

Issue #84, opened 2024-02-23
Ladder Climb-down Behavior


frendship-loost:
The ladder climb-up behavior feels nice and sticky enough. However, walking backwards in an attempt to climb down a ladder is more challenging, having to sort of fall down and air-strafe back into the ladder in order to catch it. Could this be due to the forward-facing player raycast? There is no catch behind the player to detect something like this, so I guess the awkward climb-down sort of makes sense. I don't expect this is an easy change, but it should be documented somewhere I think.

frendship-loost:
Thanks for your time, always excited for updates on Cogito!

PR #96, opened 2024-03-10
Implement intent-sensitive ladder snapping, sprinting

FailSpy:
More controller improvements!
Addresses #84

- Implemented sprinting on ladders, with exported vars to enable/disable and set sprinting speed on ladder
- `is_sprinting` will now change correctly so stamina doesn't burn whilst on the ladder for no reason
- Added intent-sensitive ladder movement. Ladder movement will generally prefer "W" being up, but if the player is pointing heavily downwards, then their "W" direction is switched to downwards, and "S" is upwards.
- If the player is looking downwards, and facing opposite the ladder as they enter its collider, it assumes the player is attempting to go down the ladder and snaps them on to it
- Additionally, players should now be able to run perpendicular to a ladder without getting caught on it, as the player now needs to be at least somewhat facing the ladder to go up it.

### DIFF
--- a/COGITO/Scripts/ladder_area.gd
+++ b/COGITO/Scripts/ladder_area.gd
@@ -1,14 +1,19 @@
 extends Area3D
 
 func _ready():
-	body_entered.connect(_on_body_entered)
+	body_shape_entered.connect(_on_body_shape_entered)
 	body_exited.connect(_on_body_exited)
 
 
-func _on_body_entered(body):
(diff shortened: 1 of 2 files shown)


Phazorknight (approved):
This is amazing!
Subjectively I feel like jumping while on the ladder has become a bit more "sticky", but I honestly don't mind. the benefits of these changes make navigating ladders way better.

Thank you!

Issue #101, opened 2024-03-11
Crash when opening escape menu


ac-arcana:
On the latest build I am crashing whenever I press escape.
It is crashing on player_state_savetime on tab_menu_controller.gd:82

I deleted my userdata and this crash persists.

Phazorknight:
Does this line appear in your output log?
`CSM: Get existing player state: No player state found for slot [slot]`

In any case, I'm pretty sure this is due to my careless implementation on how the menu reads any existing saved 😅 (deleting the data in this case unfortuantely won't help).
Should hopefully have a fix for you shortly...

ac-arcana:
![image](https://github.com/Phazorknight/Cogito/assets/5461795/5d98c70e-da38-46fc-9d16-c8bfacabdf5f)

Phazorknight:
Can you give https://github.com/Phazorknight/Cogito/commit/aca79f53e292f55dbdc8b99b821605978c86f484 a try and see if that fixes it for you?

ac-arcana:
Crashing in a different spot now...

![image](https://github.com/Phazorknight/Cogito/assets/5461795/f774a4cf-099e-41cc-afd5-e851bdbf84a6)

![image](https://github.com/Phazorknight/Cogito/assets/5461795/0e9903b2-6391-4d6a-a907-3277ed17d48e)

Phazorknight:
Okay, so this error is related to how the system time and date is read into a string and then parsed.
It just occurred to me that this string can probably differ in it's formatting depending on the OS settings.

Unfortuantely that's gonna be a bit tricky to troubleshoot for me, but if you could just add this line in the `_ready()` function of the cogito_scene_manager.gd and let me know what it prints out, it should help me narrow it down.

`print("System datetime string: ", Time.get_datetime_string_from_system(false))`

ac-arcana:
![image](https://github.com/Phazorknight/Cogito/assets/5461795/3b955b96-f887-4b4d-9480-130c36365ab0) 
as you requested :)

Phazorknight:
@ac-arcana this looks like the time string format I worked with, so still odd that you got an error.

Were you able to give @FailSpy 's PR a try?

ac-arcana:
It is fixed, thank you!

PR #103, opened 2024-03-11
Use unix timestamps instead of strings

FailSpy:
#101 
Happened to spot this out of the corner of my eye, wanted to suggest this as a potential way forward. Also allows future time formatting capability, as it merely gets stored as a Unix timestamp which is very stable output for storage, and then converted to how it wants to be formatted when it goes to render it.


FailSpy:
Force push was just to squash the commits to make it a little neater

Phazorknight (approved):
Thanks so much for this!
Gotta admit that I didn't put too much thought into saving the timestamp.

PR #111, opened 2024-03-12
Scaling viewport appropriately

FailSpy:
When you change your resolution, this PR makes it actually change the Viewport's render resolution, for both 3D and UI.
This means when you choose 1920x1080, you will actually be rendering at 1920 x 1080 (unless you change your render scale, then your UI is rendering at 1080p, but your 3D is rendering at the specified scale :P)

This includes @grgp's changes at #90 for GUI scaling, which while I think less _necessary_ now with the display resolution rendering scaling up the UI better, @grgp's GUI scale slider makes it so people can choose what size they're comfortable with.
Worth noting, the GUI scale slider will not change the rendering resolution of the UI, merely just scale up or down the pixels that are already there. Without doing a whole SubViewport thing, or greatly sacrificing 3D rendering, best I can tell this is not possible.

Addresses #81


Phazorknight (approved):
This works great! Thanks so much for this.

I actually think I have a decent solution for improving the GUI scaling but will make that a separate commit.

FailSpy:
I'll be very curious to see the solution! If you can explain it I'd also be happy to have a shot at implementing it.

Phazorknight:
I've included it in https://github.com/Phazorknight/Cogito/commit/003d035fe75bc7a915e7a6d13353f3850b646a39
Simply put it's just setting the Project's window stretch mode to "canvas_items" which improves the UI scaling behaviour.

It's not pixel perfect, noticeably reduces the aliasing that was present beforehand, making it a "good enough" option in my book.

FailSpy:
@Phazorknight `canvas_items` doesn't do what you think it does. I agree with you it *should* do UI scaling, but it doesn't. 

It will actually change the viewport render resolution to be whatever you scaled the window to, and has the canvas items render layout reflect that to be similar to the requested resolution. To demonstrate what I mean, here's a fullscreen screenshot of me allegedly running at 320x180, with the window resized to my screen resolution at 2560x1440 with the `canvas_items` feature running:

image

So really, the game is rendering at 2560x1440, despite my requested resolution at '320x180'.

Phazorknight:
Darn it. 😖  I'll reopen the issue and will undo the project stretch mode setting.

FailSpy:
@Phazorknight Yeah, I really was disappointed as well. Would've been nice for Godot to just... support pure 2D scaling like that. :sob:

PR #114, opened 2024-03-12
Create player layer; make interactions be blocked by Environment

### LINKED ISSUE #110: Interact with object through walls
Hi, in the current master, on a certain angle, it is possible to interact with objects through a wall:

lightenabler
This is the red light button
toiletenabler
and this is the toilet on the other side of the wall.

FailSpy:
This one's pretty self-explanatory. Resolves #110


FailSpy:
I was going to add an exported check to the Player controller to adjust this, but I realized it's probably better to let devs adjust the layering system themselves to suit their needs. I think this is good for a default, and if devs want something else then it's very simple to change back or make work in other ways without being too prescriptive.

Phazorknight:
Not sure if this is needed?

Issue #110 is caused by the walls not set on collision layer 2. Setting them to layer 2 fixes it, so I'd personally not change any existing layer structures on the player interaction component, especially since it would force users to update the collision layers on all their interactive objects.

FailSpy:
As it is right now, in the Project Settings of the Cogito Godot project is Layer 1 is defined as "Environment", and Layer 2 is defined as "Interactables". To me it seems conflicting to have to make all your environment pieces "Interactables" to be physically blocking interactions.

So I don't see that it would be a good idea to have users be moving all of their environment that is not literally Interactable onto the Interactables layer just to get this feature out of the environment.

All I've done in the PR is move the player on to its own layer so that way the interaction raycast can't hit it, and include Environment as a collision layer for the raycast to have environment block interactions.

Unless this layer setup is no longer of interest, at which point the layer names should probably be changed. Though personally, I would prefer environmental pieces on the default layer 1 to be blocking interactions.

Also by not encouraging the separation of these layers, then you would have no way to discern Interactables vs Environment purely via Raycast if you did want interactions to work through walls, you have to do the raycast, then check if what it hit is in the group.

Phazorknight:
That makes sense!
Having the environment block by default would definitely be great.
Let me take a closer look at the PR.

Phazorknight:
Hey @FailSpy , when I open this on my end, playtesting the project makes my resolution all pixely, no matte which one I select from the options + the temp UI scaling slider is gone.

Could you take a look to undo any changes to those in this PR?

I'd try to cherrypick, but with the update layersettings I'd like to keep the project.godot file from this PR.

FailSpy:
@Phazorknight give it a try now?

FailSpy:
Oh wait, hold off. Found a major issue with the layering setup: bunch of stuff now doesn't collide with the player. Attempting a workaround that minimizes breakage of other people's existing work...

FailSpy:
@Phazorknight 
Okay, so I've gotten rid of the player layer to be minimally conflicting with existing projects as it'd be a nightmare. Encountered all sorts of interesting bugs. As nice as it would be to have a separate player layer, I feel it's a bit much to do right now, so this is the absolute minimum set of changes to accomplish this PR's goal

- Environment layer is included in Interaction raycast
- Player class when setting up the Player Interaction Component manually configures the interaction raycast to *exclude* the Player (and thusly, its colliders) specifically, that way the player can be on *any* layer, and the interaction raycast will ignore it.

COGITO/Components/PlayerInteractionComponent.gd
```diff
@@ -269,7 +269,7 @@ func send_hint(hint_icon,hint_text):
 
 # This gets a world space collision point of whatever the camera is pointed at, depending on the equipped wieldable range.
 func Get_Camera_Collision() -> Vector3:
-	var viewport = get_viewport().get_size()
+	var viewport = get_viewport().get_visible_rect().size
```
FailSpy, line 272: I've noticed that if I have my game window scaled differently from the selected resolution, camera raycasts bug out and guns will fire off in weird directions. Turns out this has been an issue regardless of scale mode after some testing. This fixes that

FailSpy:
@Phazorknight Okay, properly ready now to be pulled in. Give it a shot, see how it goes for you.

Phazorknight (approved):
Seemed good on my end! Gonna approve.

Thank you so much for this!

PR #115, opened 2024-03-13
Added a toggle for Bunnyhopping

ksjfhor:
not only the toggle, but also a few types added to variables


Phazorknight (approved):
Sweet, thanks for this!

PR #117, opened 2024-03-13
Ladder collision fixes and buffer timer

FailSpy:
Whilst doing the last-minute fixes on PR #114, I realized there's a couple of issues with the ladder system that while avoidable with good collision shape placement, it's probably easier to implement these fixes:

- When stepping on the ladder, you have a half second cooldown (configurable) before checking if you're on the ground and you should "step off" the ladder. This makes it so if you're just stepping on to the ladder, you can't accidentally register as "on ground" and immediately come off it. The idea is the player will move within that .5 seconds, and the ground check will have changed.
- In the ladder class, you can configure a "ladder collision" object. If this is set, it will disable the collision object whilst the player is on the ladder, and then set it back to whatever it was when the player comes off it. This prevents the collision object from blocking the player from going up/down the ladder if the player's collider isn't snapped perfectly.

Made this as a separate PR just to make sure these changes get their own chance for review.

You can see the issues with this on `main` by attempting to go down the ladder in the `COGITO_01_Demo` scene, which is fixed in this PR.

### DIFF
--- a/COGITO/Scripts/player.gd
+++ b/COGITO/Scripts/player.gd
@@ -103,7 +103,9 @@ var on_ladder : bool = false
 @export var CAN_SPRINT_ON_LADDER = false
 @export var LADDER_SPEED : float = 2.0
 @export var LADDER_SPRINT_SPEED : float = 3.3
-const LADDER_JUMP_SCALE : float = 0.5
+@export var LADDER_COOLDOWN : float = 0.5
+const LADDER_JUMP_SCALE = 0.5
+var ladder_on_cooldown = false
 
 @export_group("Gamepad Properties")
 @export var JOY_DEADZONE : float = 0.25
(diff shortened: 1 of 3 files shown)


Phazorknight (changes requested):
player.gd: 
Could the variables that were made explicitly typed be added back in? This might've just been an overlap of your PR as well as the other PR that added explicit types to player.gd

line 107: float
line 108: bool

Otherwise ladder behavior feels definitely better and snappier, especially when climbing down!

COGITO/Scripts/player.gd
```diff
@@ -100,12 +100,12 @@ const WALL_MARGIN : float = 0.001
...
 var on_ladder : bool = false
-@export var CAN_SPRINT_ON_LADDER = false
+@export var CAN_SPRINT_ON_LADDER : bool = false
 @export var LADDER_SPEED : float = 2.0
 @export var LADDER_SPRINT_SPEED : float = 3.3
 @export var LADDER_COOLDOWN : float = 0.5
-const LADDER_JUMP_SCALE = 0.5
-var ladder_on_cooldown = false
+const LADDER_JUMP_SCALE : float = 0.5
+var ladder_on_cooldown : bool = false
```
FailSpy, line 108: :+1:

Phazorknight (approved):
Thanks! looks good!

Issue #108, opened 2024-03-12
Carryable Object Destruction Results in Error


ac-arcana:
If you are carrying a carryable and it is destroyed while you are carrying it, then when you next press the carry input there is an error.

Phazorknight:
Thanks for flagging this.
Have a pretty good idea on what causes this and how to fix it, so should be able to get this done today.

Phazorknight:
Please check https://github.com/Phazorknight/Cogito/commit/003d035fe75bc7a915e7a6d13353f3850b646a39.

This adds a check to the Player Interaction component to see if the carryable object reference is still valid, and if not clears it.
You should no longer get crashes, but it's possible that you have to press an interact button first before the reference is updated. Will see if I can somehow make it cleaner, but if it works for you, I'd consider this resolved.

ac-arcana:
Solved!

PR #118, opened 2024-03-13
Add automatic leave() for deleted carried objects

FailSpy:
Addresses #108

You mentioned you need the reference to be cleaned up in the current implementation, thought I'd throw this patch in that makes it so a Carryable object will call 'leave()' as part of its cleanup on destruction.


Phazorknight (approved):
You're on fire today. 🔥 

Also damn, i didn't know about _enter_tree and _exit_tree, those give me all kinds of ideas on how to make the overall code better. 🤔

PR #124, opened 2024-03-19
Basic light detection

Nutcrawler:
Added simple camera rendering a light-only debug draw, and set the visibility attribute based on the luminance of the texture


Phazorknight:
Hi, thanks for this!
Took me a moment to understand what you're doing. I think this is a decent way to get a more accurate light-level of where the player is at, but it's also very hard to control the value. Still, could be pretty useful for some!

For me to commit this into main, I'd request the following changes:
- Make this its own attribute class based on a CogitoAttribute (maybe call it something like Lightmeter / Lightlevel?) as to not overwrite current visibility attribute.
- Move all of it's code out of the Player.gd and all nodes out of the Player.tscn. Would probably make sense to instead have them in the Attribute PackedScene.
- Include a documentation paragraph that I can add into the Documentation.

Thanks!

PR #125, opened 2024-03-24
Sliding Sound

ac-arcana:
Added a sound to be played when crouch sliding.


Phazorknight:
Simple but pretty neat!

I think in the current implementation the directionality is too strong, might be better to just make the sound "2D" or use the quick audio add-on to play/spawn it.
The volume level also seemed quite inconsistent, one slide it was okay, the next slide it was very quiet.

ac-arcana:
I want to stay away from 2D sound mainly because I want to support multiplayer with my project.

I didn't notice an inconsistent volume but I'll see if I can reproduce it and come up with a solution.

The quick audio might not work well because it wouldn't move the sound source with the player, I think. I might give it a try and test that theory, though.

ac-arcana:
Okay, changed the play method to Quick Sound. You can keep a reference to a node and reparent it, so this allows the sound to follow the player. I have no idea *why* the volume was not consistent on the audio stream player and is consistent with this, but oh well. Seems to have fixed the consistency problem.

I would like to note that the Sound Effects volume setting currently seems to have no effect on these sounds. I was checking because I was curious if Quick Audio was adding SFX to the SFX bus.

ac-arcana:
As a bonus, since I basically just did what is in a "todo" tag, the part saying that the walk / sprint threshold needed to be made to variables started to bug me, so I've made those variables, too. I switched from length() to length_squared() on velocity to save a slight bit of performance

Phazorknight:
I just gave this another try, and I am getting some errors:

image

This happened both when testing in the Lobby and Laboratory demo scene.

ac-arcana:
Ah I didn't notice that. I think it's a problem with reparent. I probably need to defer the reparent call. I'll look at this some time this week to fix it

ac-arcana:
I believed I've fixed the error.
Also removed some other prints I left in when I was debugging

Phazorknight (approved):
Looks good!
Thanks so much!

PR #132, opened 2024-04-01
Correct spelling mistakes in Lobby scene hints

samaritan-sw:
Hi, first of all - very cool project! :)

My humble contribution: I've found some spelling mistakes across demo hints in the lobby scene, so I've cleaned them up, rephrased some small bits, and split up some paragraphs in a way that makes a bit more sense visually.

Cheers!

### DIFF
--- a/COGITO/DemoScenes/COGITO_04_Demo_Lobby.tscn
+++ b/COGITO/DemoScenes/COGITO_04_Demo_Lobby.tscn
@@ -7433,18 +7433,6 @@ surface_material_override/0 = SubResource("StandardMaterial3D_yhahm")
 [node name="Hint_01_Welcome" parent="DEMO_HINTS" instance=ExtResource("62_y3og2")]
 transform = Transform3D(0, 1, 0, -1, 0, 0, 0, 0, 1, 1.90368, 1.44587, -15.4)
 
-[node name="ReadableComponent" parent="DEMO_HINTS/Hint_01_Welcome" index="6"]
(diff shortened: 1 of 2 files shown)


inertaert (commented):
I'm just a random guy on GitHub, but I agree that these small changes greatly improve readability and think @Phazorknight should approve them.

One minor note: The addition made at #L7436-L7447 should probably be made directly to the default text in [DemoPrefabs/cogito_demo_hint.tscm](https://github.com/Phazorknight/Cogito/blob/528ed415ce5a7c2f52ec357a2da84458ece6a65c/COGITO/DemoScenes/DemoPrefabs/cogito_demo_hint.tscn) instead: https://github.com/Phazorknight/Cogito/blob/528ed415ce5a7c2f52ec357a2da84458ece6a65c/COGITO/DemoScenes/DemoPrefabs/cogito_demo_hint.tscn#L76-L85

Nice work.

Best regards.

samaritan-sw:
Hi @inertaert, thanks for your feedback!

You're right, I didn't realise the demo hint had its Readable Content property set inside its own scene - that should be correct now.

Phazorknight (approved):
Thank you so much for these!

I can only use "english isn't my native language" for so long, really need to run stuff through a grammar and spell check before committing. 😅

samaritan-sw:
@Phazorknight I'll tell you a little secret...it's not my native language either! 😅 So take my submissions with a grain of salt...

PR #133, opened 2024-04-01
Fix COGITO_O1_Demo.tscn scene load error

samaritan-sw:
Hi,

On the current main branch I'm getting an error trying to load COGITO_01_Demo.tscn due to the chest inventory resource being referenced but not actually defined anywhere in the scene file (probably due to some mixup during a commit).

I've defined a new InventoryPD resource for the chest wit 4 slots, 2 potions and 2 batteries (which seems to be what it contained originally), and the scene now loads properly. It seems resource ids in scene files are constantly redefined by Godot, so sorry if this PR contains more line changes than strictly necessary!


ac-arcana:
I can confirm I can't load that scene either.

image

PR #135, opened 2024-04-01
Fix for sliding sound when jumping mid slide

ac-arcana:
This commit fixes the slide audio playing mid jump and should fix the crash if the slide audio player instance was lost.


Phazorknight (approved):
Huzzah! Works great, thank you!

PR #139, opened 2024-04-02
Add "is_carryable_while_wielding" property

samaritan-sw:
I find it a little tedious having to put away my weapon in order to carry something as small as a battery or a potion. I think it would be better to leave devs with an easy option to disable this for lighter items or across all items entirely.

To that end, I've implemented a really simple bool property in CarryableComponent.gd which lets you toggle this behaviour. Of course, in what I believe to be the spirit of the plugin, I think we can let users worry about the intricacies of their own implementations (in terms of handling potential collision issues between the viewmodel and the carried object if the carry distance is really close for example).

I've tested this and have found no immediate issues but maybe there is a technical reason why this restriction was in place?


Phazorknight (approved):
Thanks for this!
I think this implementation makes the most sense, enabling devs to define which objects can be carried while wielding.

PR #141, opened 2024-04-02
Instantiate nodes using position instead of global_position

samaritan-sw:
This fixes an issue where every time a node is set to spawn upon another's death, or a spawner is used, a non-fatal error message "Condition !is_inside_tree is true" pops up.

Repro steps: shoot a target in any scene and the error will pop up in the console.

I believe this is due to adding the child to the tree before setting its global_position property. Unfortunately, setting the global_position before adding the child to the tree seems to prevent the node from showing (or at least from spawning in the right position).

Using "position" to instantiate the nodes seems to fix this, and I've observed no change in behaviour - however this solution might not follow best practice, so people more familiar with the engine should feel free to come up with something better if needed.


Phazorknight (approved):
Thanks for this!
I'm very happy to finally get rid of these messages.
Don't really notice a difference either, I guess we'll see if there's weird behaviour down the line.

PR #142, opened 2024-04-02
Extend particle lifetime to avoid render server error

samaritan-sw:
Tiny fix to prevent Godot throwing render server error "Condition "p_length < 0.01" is true" every time simple_particle_puff is emitted.


Phazorknight (approved):
I was wondering for how to fix this, thanks so much!

PR #143, opened 2024-04-03
Player script style changes

ac-arcana:
Added a class name for the player script.
Moved the variables that were hidden between functions to the top. Variables proceed as export, public, onready.
Code region added for variables.
Spacing between functions is two lines.
Added documentation describing class to beginning.


Phazorknight (approved):
Thanks so much! Appreciate the spring-cleaning. 😁

PR #145, opened 2024-04-03
Fix laser angle when shooting close to walls

samaritan-sw:
Bullet direction not necessary for drawing the laser between rifle and point of impact - was causing lasers to draw perpendicularly to the player when close to walls.

Old:
image

New:
image


Phazorknight (approved):
Thanks a bunch!

PR #146, opened 2024-04-04
Pause Menu Style Changes

ac-arcana:
Added a class name so devs can easily override the menu with their own custom submenus (I do this in my MP branch)
Changed variable order to export, public, onready
Made variable region
Changed spacing to 2 lines between functions.

There was an old region for UI Audio, but when I collapsed it the region didn't work properly. It stretched from the UI variables at the top to the end of the audio playing function, which didn't make much sense to me.


Phazorknight (approved):
Looks good, thanks for this!

Issue #144, opened 2024-04-03
Create new Wieldable


ksjfhor:
**Cogito and Godot Engine Version:**
Cogito: beta 202403.25
Godot: 4.2.1.stable

**Description:**
I am too stupid to create a new wieldable, I tryed it on my own, until I found your YT-Video from yesterday in the docs, so I started fresh with the Video, from around 2 min until 5 min ( I watched the whole Video
But there must be one thing that I miss, because the darts does not get marked as pickupable .. :/

**Reproduction steps:**
Please see the screenshots what I did to the nodes.
I basicly took a look on the laser rifle for the wieldable and the battery as ammo.
I created and saved the InventorySlot for the dart and for the animations.
Same for adding the Dart to the player and the PlayerInteractionComponent.

**Expected behavior:**
I expect that I could pick up the darts and make them a wieldable to throw the dart.

**Screenshots:**
![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/45621a8c-9c3a-4630-a2b9-2402dca6a417)
![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/7f1f4523-e816-4f5c-b070-b962adf07636)
![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/b8cb67a6-bbb8-4d08-9754-d36051cbac39)

![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/24fae4be-9f14-4da3-88db-d99ce3c65826)
![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/ddecf25e-4dd3-4e39-b7f4-53df28345a67)

![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/b7dd3575-5d39-4f7b-8ce3-7ee0108c98de)
![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/7ff32a55-211d-4236-bfa9-dd2fe8a9db99)

![grafik](https://github.com/Phazorknight/Cogito/assets/87974401/a24eb957-b652-4501-806f-59802ea27e6a)

BTW: Line 7 in the InventoryItemPD.gd is "despcription" not "description" ;)
And I got confused by the CogitoClass-node, the node of the battery is a Rigidbody3D but the script extends a Node3D, is there a reason why ?

Phazorknight:
Firstly, sorry that this is giving you trouble. I'm sure we can get this to work though.

[...]

PR #147, opened 2024-04-04
Fix typos in "description" property of pickup items

samaritan-sw:
- Changed "descpription" field name in InventoryItemPD.gd to "description" (mentioned in #144) 
- Fixed some typos in pickup item descriptions
- Also a tiny fix to the laser rifle shoot animation which was only playing every other shot. Shortened the animation by 0.1 seconds so the animation plays correctly every time the rifle is shot.


Phazorknight (approved):
You're the real mvp.

PR #151, opened 2024-04-06
Enhancement/wieldable decouple

### LINKED ISSUE #150: Decouple Wieldables from player scene
As a general enhancement, I'd suggest making it not required for the wieldable scenes to be pre-instantiated in the player scene:
image

Making a few changes this should be pretty simple to do and would ease the process of creating new wieldables.

I'll submit a proof of concept PR showing how I would personally go about this.

brian-holsters:
fixes #150

### DIFF
--- a/COGITO/Wieldables/Wieldable_Throwable.gd
+++ b/COGITO/Wieldables/Wieldable_Throwable.gd
@@ -2,6 +2,8 @@ extends CogitoWieldable
 
 
 @export_group("Throwable Settings")
+## Leave empty if projectile and pickup are the same scene
+@export var projectile_override: PackedScene
 ## Speed the projectile spawns with
 @export var projectile_velocity : float
 ## Node the projectile spawns at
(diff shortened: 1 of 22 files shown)


COGITO/Wieldables/Wieldable_Throwable.gd
Phazorknight, line 1: Projectile spawning:
I would not default to the projectile prefab being the same as the pickup item.
I can see cases where devs might want a the spawned projectile be a different scene than the pickup-item
(items which thrown state would differ from their pickup item state, something like armed grenades for example).

Phazorknight (changes requested):
Requested a change to Throwables defaulting to the pickup scene as the projectile. otherwise this looks solid.

COGITO/Wieldables/Wieldable_Throwable.gd
brian-holsters: solved by adding an optional export to the wieldable script, here's a video of it working:

https://streamable.com/i9i99n

Phazorknight (approved):

PR #153, opened 2024-04-06
item charge no longer shows if item has the no_reload tag

### LINKED ISSUE #152: Throwable item "ammo count" is inconsistent with how stack size and ammo appears on other items
image

The dart item has both an "ammo counter" which is fixed at 1 and a stack size counter which shows how many you actually have.

I suggest removing the ammo counter to make it consistent with other stackable items.

brian-holsters:
Fixes #152 

Before:
image

After:
image


Phazorknight (approved):
Awesome, was about to fix this myself this weekend but you were faster.
Appreciate it!

PR #157, opened 2024-04-07
Base Tab Menu

ac-arcana:
Separate out the basic tab menu input function and nodes_to_focus to another class. OptionsTabMenu now extends this class. This will make it easier to create other tab based menus with controller support in the future.

Also some minor style changes in OptionsTabMenu.


Phazorknight (approved):
Thanks! Looks good.
Will definitely be helpful for future updates.

PR #161, opened 2024-04-13
Footstep sounds in sync with head wobble, wobble intensity is adjustable

BrodyB:
Started working on a game that uses Cogito as a base, and these seemed like good general enhancements I could bring back in.
1. Footstep sounds no longer use a timer and instead trigger when the head wobble sine value goes over 0.9, ensuring headbob and footsteps are in sync.
2. Headbob intensity values are exposed, and adjusting the headbob settings scale the intensity only. This way footsteps have a consistent interval across settings.
3. Headbob settings are now Full (100%), Average (70%), and Minimal (10%). Though maybe there should be an Off setting (0%)?


Phazorknight (approved):
Thanks for this!
Definitely a nice improvement.

PR #162, opened 2024-04-14
Cogito_Switch style changes and Player Hud Style Changes

ac-arcana:
Cogito_Switch:
Added class name
Reordered variables
Added variable code region
Double spaced functions

I needed a class name so that I could find all switches in a scene in order to synchronize their state across multiplayer clients.

### DIFF
--- a/COGITO/CogitoObjects/Cogito_Switch.gd
+++ b/COGITO/CogitoObjects/Cogito_Switch.gd
@@ -1,11 +1,12 @@
 @icon("res://COGITO/Assets/Graphics/Editor/CogitoNodeIcon.svg")
+class_name CogitoSwitch
 extends Node3D
 
 signal object_state_updated(interaction_text: String) #used to display correct interaction prompts
 signal switched(is_on: bool)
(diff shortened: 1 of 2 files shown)


ac-arcana:
For player HUD:
Added class name
Reordered variables
Added variable region
Moved setup of player and attributes to its own function - this is so that if the player object is changed the HUD can be reset to the new player.
Removed any children from the ui_attribute_area to clear any UI from any previous player.

The reason I moved player and inventory setup to its own function is because I am replacing the player in multiplayer. However, this may also be use for games with respawn mechanics, or squads where you can switch members.

ac-arcana:
Just improved player_hud a little more, by creating a helper function so that you can pass a player into the function directly.
Also calling delete_interaction_prompts to pervent stuck prompts from any old players.

I think I'm done making continued changes.
Sorry, I realize this is a bad commit as I have done style changes AND made some changes to the hud. It seems further commits just get rolled into my previous pull request...

Phazorknight (approved):
Thanks! Approved.

PR #163, opened 2024-04-14
Cogito Door Style Changes plus Close Door Without Interactor

ac-arcana:
Added a class name.
Reordered variables
Created a variable region
Double spaced functions
Added a null check on player_interaction_component in close_door. This is so that entities that are not the local player can close doors without a crash


Phazorknight (approved):
Nice! Thanks so much for this!

PR #165, opened 2024-04-16
refactor: Store player attributes in Dictionary instead of Array

aronand:
Player attributes are accessed by name, thus storing them in a Dictionary makes more sense than in an Array.

These changes edit attribute storing, saving, and loading to use Dictionaries instead. An attempt is also made to make the code ever so slightly more readable by using more descriptive variable names where applicable.

Changes are tested in test maps and seem to function as intended.


Phazorknight (approved):
Sweet, thank you for this!
Approved.

PR #169, opened 2024-04-18
Simplifies interaction input handling and refactors reload logic to be more performant

aronand:
- Simplifies PlayerInteractionComponent by merging input checks for interact and interact2 to a single block
- Removes a RayCast3D.is_colliding() check and gets that information via RayCast3D.get_collider() (null if no collision)
- Rewrites the reload logic (removes nesting, removes the while loop, makes the for loop faster by removing the use of InvetoryPD.remove_item_from_stack())

Reload time before:
reload_before_self

Reload time after:
reload_after_self

### DIFF
--- a/COGITO/Components/PlayerInteractionComponent.gd
+++ b/COGITO/Components/PlayerInteractionComponent.gd
@@ -78,47 +78,29 @@ func interactive_object_exit():
 	object_detected = false
 
 
-func _input(event):
-	if event.is_action_pressed("interact"):
-		
+func _input(event: InputEvent) -> void:
+	if event.is_action_pressed("interact") or event.is_action_pressed("interact2"):
+		var action: String = "interact" if event.is_action_pressed("interact") else "interact2"
+
 		# if carrying an object, drop it.
-		if is_carrying and is_instance_valid(carried_object) and carried_object.input_map_action == "interact":
+		if is_carrying and is_instance_valid(carried_object) and carried_object.input_map_action == action:
 			carried_object.throw(throw_power)
 		elif is_carrying and !is_instance_valid(carried_object):
 			stop_carrying()
-		
+
 		# Checks if raycast is hitting an interactable object that has an interaction for this input action.
-		if interaction_raycast.is_colliding():
-			interactable = interaction_raycast.get_collider()
-			if interactable.is_in_group("interactable"):
-				for node in interactable.interaction_nodes:
-					if node.input_map_action == "interact":
-						node.interact(self)
+		# This could be run each frame instead and the key be checked when finding an interactable,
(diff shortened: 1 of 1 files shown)


COGITO/Components/PlayerInteractionComponent.gd
```diff
@@ -78,47 +78,29 @@ func interactive_object_exit():
...
-						node.interact(self)
-		interactive_object_exit()
-		if is_wielding:
-			equipped_wieldable_item.update_wieldable_data(self)
-	
-	
-	if event.is_action_pressed("interact2"):
+func _input(event: InputEvent) -> void:
+	if event.is_action_pressed("interact") or event.is_action_pressed("interact2"):
+		var action: String = "interact" if event.is_action_pressed("interact") else "interact2"
```
aronand, line 83: Input and InputEvent don't seem to have a way to return the action name, thus a bit hacky way of getting the action name. Needs a different solution if more interaction keys are introduced, or moving the interaction check to Node._process() instead.

COGITO/Components/PlayerInteractionComponent.gd
```diff
@@ -222,51 +204,52 @@ func attempt_action_secondary(is_released:bool):
...
-	if !equipped_wieldable_node.animation_player.is_playing(): #Make sure reload isn't interrupting another animation.
-		equipped_wieldable_node.reload()
-		
-		while ammo_needed > 0:
-			if equipped_wieldable_item.get_item_amount_in_inventory(equipped_wieldable_item.ammo_item_name) <=0:
-				print("No more ammo in inventory.")
-				break
-			for slot in inventory.inventory_slots:
-				if slot != null and slot.inventory_item.name == equipped_wieldable_item.ammo_item_name and ammo_needed > 0:
-					inventory.remove_item_from_stack(slot)
```
aronand, line 257: InventoryPD.remove_item_from_stack() consumes one item at a time, making this loop take longer than is necessary. A possible refactor to it would be to allow the user to define the amount to remove as well.

Phazorknight (changes requested):
Overall this does some nice cleanups.

There's an issue though with the changes to the reload: It seems like the reload_amount properties of AmmoItemPD is not considered, thus a battery is just adding 1 to the charge of the flashlight. Or the clip of the Laser Rifle is just adding 1 to the Laser Rifle charge.

Would you be able to fix this?

aronand:
Good catch! I'll have a look during the weekend.

aronand:
Fix is in now. Tested both the flashlight and the laser rifle, which now seem to function as intended.

Phazorknight (approved):
Looks good, thanks!

PR #171, opened 2024-04-20
when loading a player state, unequip current wieldable before loading

### LINKED ISSUE #166: Wieldable nodes/models not clearing when loading a save that's within the same scene.
Godot 4.2.1.stable | Cogito beta 202404.01

**Description:**
When I load a save state that's within the same level scene the player is currently in while having a wieldable equipped. That wieldable mesh stays active while the wieldable I had equipped in the save state is also being equipped, resulting in doubled-up wieldables in the viewport.

**Reproduction steps:**
1. Start a level scene, pick up a wieldable and equip it.
2. Save the game (while wieldable is equipped/being held)
3. load the game while still being in the same level scene and still holding a wieldable
4. When loading is done, the wieldable that was being held when loading is still there, while the wieldable that was being held in the save state is getting equipped.

**Expected behavior:**
Equipped wieldables should be cleared when a game is loaded / before the player state is being loaded/instantiated.

Tagging @brian-holsters

brian-holsters:
Fixes #166 

Added an optional argument to change_wieldable_to which allows to skip the await. 
This is required for quick loading the current scene since otherwise the await messes up the loading time resulting in unequipping the wieldable from the loaded data instead of the old one.

### DIFF
--- a/COGITO/Components/PlayerInteractionComponent.gd
+++ b/COGITO/Components/PlayerInteractionComponent.gd
@@ -183,13 +183,13 @@ func equip_wieldable(wieldable_item:WieldableItemPD):
 
 
 
-func change_wieldable_to(next_wieldable: InventoryItemPD):
+func change_wieldable_to(next_wieldable: InventoryItemPD, force_immediate := false):
 	is_changing_wieldables = true
 	if equipped_wieldable_item != null:
 		equipped_wieldable_item.is_being_wielded = false
 		if equipped_wieldable_node != null:
 			equipped_wieldable_node.unequip()
-			if equipped_wieldable_node.animation_player.is_playing(): #Wait until unequip animation finishes.
+			if not force_immediate and equipped_wieldable_node.animation_player.is_playing(): #Wait until unequip animation finishes.
 				await get_tree().create_timer(equipped_wieldable_node.animation_player.current_animation_length).timeout 
 	equipped_wieldable_item = null
 	if equipped_wieldable_node != null:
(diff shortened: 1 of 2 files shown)


aronand:
Your fix seems to work as intended. However, while testing it I discovered a bug with the load system, maybe someone can try to repro and verify if it's not just me somehow (tried with a clean workspace from main as well):

1. Make a save before picking up a wieldable
2. Pick up wieldable
3. Wield wieldable
4. Load save
5. Notice that the wieldable items is not appering the world where it's supposed to

aronand:
One bug I did notice upon further testing is that the UI still retains information about the wieldable after loading a save:
image

brian-holsters:
After further testing, I don't think this is ready to merge...

These are the bugs related to loading with equippables so far:

- When loading a save where you had no wieldable equipped while having a wieldable equipped, the HUD doesn't update.
- When loading a save where you had the same wieldable equipped or you come from another scene, the charge HUD dissapears and trying to unequip the wieldable just re-equips it.

Phazorknight:
So I've attempted to fix this myself in https://github.com/Phazorknight/Cogito/commit/85401bcd1f4b1dbb3d89a6503722fb4a8c126592

The approach I took is that I don't attempt to do anything before loading, but instead try to reset all leftover references when the PlayerInteractionComponent is loaded (which is when `set_state()` is called) and then equip the wieldables by just calling `use` on the wieldable item that was equipped when the save state was created.

Phazorknight:
Closing as this is currently not an issue in `main`

PR #173, opened 2024-04-20
Separates interactable updating from PlayerInteractionComponent to InteractionRayCast

### LINKED ISSUE #170: Separate interactable detection from PlayerInteractionComponent
Currently PlayerInteractionComponent handles both player interactions and detecting interactables in the player's view.

I propose that the detection functionality is decoupled from the controller and given to the RayCast3D that the Player holds. This node will then send signals whenever an interactable comes into view or leaves the view, to which the controller will respond to accordingly.

I've done some preliminary work and testing on it here: https://github.com/aronand/Cogito/tree/InteractionRayCast

Currently the implementation is mostly functional, with only double interaction component seemingly not working. Due to the way that the UI prompts work however, I currently have to hack them by updating them each frame, and to fix this, changes will most likely also impact the UI.

aronand:
This PR holds multiple changes, the most notable being:
- Adds InteractionRayCast that is attached to the player's interaction RayCast3D
  - InteractionRayCast is responsible for updating the currently seen interactable, if any. This is done via signals that the PlayerInteractionComponent listens to.
    - The state itself only changes whenever a new interactable comes into view, or the current interactable goes out of view
- Reorganizes UI prompt updates to only happen once, upon a state being changed
- Refactors to PlayerInteractionComponent:
  - Unused/unnecessary variables removed
  - Getters and setters added for variables to which it makes sense (e.g. state change warrants emitting a signal)
  - Removes most direct calls to interaction_raycast, instead relying on the signals
- Small simplifications + whitespace fixes where noticed (hide whitespace changes in GitHub to only focus on code changes)

Closes #170 

Depends on #169 

(Due to there being lots of atomic commits, it may make sense to squash the commits depending on how you wish the repo history to look)

### DIFF
--- a/COGITO/CogitoObjects/Cogito_Door.gd
+++ b/COGITO/CogitoObjects/Cogito_Door.gd
@@ -222,8 +222,6 @@ func close_door(_interactor: Node3D):
 		tween_door.tween_property(self,"position", closed_position, door_speed)
 	is_open = false
 	interaction_text = interaction_text_when_closed
-	if player_interaction_component:
-		player_interaction_component.nothing_detected.emit() # Forces a re-detection of the interaction prompt.
 	object_state_updated.emit(interaction_text)
 	door_state_changed.emit(false)
 
--- a/COGITO/CogitoObjects/Cogito_Door.gd
+++ b/COGITO/CogitoObjects/Cogito_Door.gd
@@ -223,7 +223,7 @@ func close_door(_interactor: Node3D):
 	is_open = false
 	interaction_text = interaction_text_when_closed
 	if player_interaction_component:
-		player_interaction_component.interactive_object_exit() #Froces a re-detection of the interaction prompt.
+		player_interaction_component.nothing_detected.emit() # Forces a re-detection of the interaction prompt.
 	object_state_updated.emit(interaction_text)
 	door_state_changed.emit(false)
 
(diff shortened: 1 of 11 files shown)


aronand:
Regarding the drop prompt, you can actually observe it being broken in main as well at the moment. I suspect that could be fixed by actually adding state to the UI, possibly `Player_Hud_Manager`, which manages it's own visibility reactively.

aronand:
After creating a simplified version of the raycast for my own project I noticed that it *and* this are spamming the unseen signal. It's a really hard one to fix though, as `is_instance_valid()` will be true for both null and for a freed object => seemingly no easy way for an early exit.

Checking for `Object.is_queued_for_deletion()` also doesn't seem to help unfortunately. I'm possibly still missing something here, because in one of my attempts things seemed to work fine. In that one I had prints in different places which possibly slowed things down enough for everything to work as intended.

Thus this is currently an imperfect solution, as I haven't come up with a way to stop the signaling.

aronand:
A modified version of the simplified raycast is now in.

One oddity I still noticed is that at least in Cogito, the freed object handling sometimes gets triggered even when only quickly hovering the cursor over multiple interactables, but in practice this seems to have no significant impact.

aronand:
The drop prompt hack is now removed as well. This is achieved by not emitting interactable signals while an object is being carried.

With that I'd say the PR is feature complete. Let me know if you spot any bugs I may have missed.

Phazorknight (approved):
Very nice. Cleans up quite a few lines.
Thank you!

PR #177, opened 2024-04-23
Add lock_state_changed signal to CogitoDoor

ac-arcana:
Just adding and emitting a signal when the door's lock state is changed. 
I used this in order to sync door locks in my Coop project, but it could be used to hook anything to a door being unlocked.

### DIFF
--- a/COGITO/CogitoObjects/Cogito_Door.gd
+++ b/COGITO/CogitoObjects/Cogito_Door.gd
@@ -4,6 +4,7 @@ extends Node3D
 
 signal object_state_updated(interaction_text:String)
 signal door_state_changed(is_open:bool)
+signal lock_state_changed(is_locked:bool)
 signal damage_received(damage_value:float)
 
 #region Variables
@@ -160,6 +161,7 @@ func unlock_door():
 	is_locked = false
 	interaction_text = interaction_text_when_closed	
 	object_state_updated.emit(interaction_text)
+	lock_state_changed.emit(is_locked)
 
 
 func open_door(interactor: Node3D):


aronand:
One tiny change could be to add a setter for is_locked, which handles emitting the signal whenever the state changes. This way no matter what causes the state to change (e.g. a lock_door() method is introduced later down the line, or someone manually sets the state somewhere else), the signal gets emitted.

aronand:
One tiny change could be to add a setter for is_locked, which handles emitting the signal whenever the state changes. This way no matter what causes the state to change (e.g. a lock_door() method is introduced later down the line, or someone manually sets the state somewhere else), the signal gets emitted.

This would look something like:
```gdscript
is_locked: bool:
    set = _set_is_locked

# ...

func _set_is_locked(value: bool) -> void:
    is_locked = value
    lock_state_changed.emit(is_locked)
```

ac-arcana:
I'm still learning gdscript. Is defining the variable like that going to call the function whenever the variable is changed, regardless of how it is changed?

I wasn't aware of setters like this in gdscript.

ac-arcana:
Thanks for linking the docs.
I see value in a setter and getter, but it also begs the question if the same should be done for is_open.
Keep in mind if doing it for both it doesn't really increase functionality for games.
I propose instead that I write a simple lock_door function. This would actually increase functionality for games and cover all use cases for locks.

aronand:
True, `is_open` could have one too. It's really just a way to ensure that something happens when the state changes, even if the state is changed in the "wrong" way (e.g. for whatever reason directly altering the variable instead of using a method in some script). This way if you also do end up having more methods altering the state and sending signals, you keep the code a bit more DRY.

ac-arcana:
The way you put DRY in all caps makes me think this is a programming
concept I haven't heard of, too. Lol

aronand:
Haha, yeah, it is 😄 DRY = Don't Repeat Yourself

ac-arcana:
I think I'm going to leave this pull request as is.

@aronand I agree with the suggestions you are making, but in practice it's a little finnicky. I feel like keeping open_door, close_door, and unlock_door public and the is_open and is_locked variables "private" would be the proper way to do this.
However, the variables are used initially to set the state for the start of the scene so they aren't entirely "private"....

Trying to do it the other way, where I use setters and getters I run into a recursion problem. The interact function is already calling functions like open_door, close_door, and unlock_door. This could be rewritten to set the variable instead, but then there is the problem of passing the interactor to the open_door function.

I just don't feel like any solution I have come up with in practice is cleaner or better, there is always something still wrong with it. If someone else wants to improve this further, feel free.

aronand:
Yeah the `is_open` and `is_locked` flags are something that are good to be publicly readable at least, as someone might want to check those in a script of their own.

I can have a go at a possible implementation during the week when I have the time, but I think it can also be turned into an issue of its own, especially as it's looking to be a larger refactor than just a small edit in logic in a single file.

Phazorknight (approved):
Simple enough and makes sense to have a singal for this.

PR #188, opened 2024-05-05
CarryableComponent Improvements

ac-arcana:
Some changes I did for my Online Coop project that could benefit Cogito:

Added the Cogito prefix to the class name
Added signals for when the carry state is changed and when it is thrown. Also changed some spacing.
Removed some repeated multiplication in the throw function.


Phazorknight (approved):
Looks all good, though technically none of the InteractionComponents have the Cogito prefix in their names so far. 😅 
Guess we could start renaming all of them then...

PR #191, opened 2024-05-07
Setup read the docs

ac-arcana:
Please don't merge this yet.
Apparently Read The Docs can compile documentation from pull requests.
I am just setting up a config file here and I'd like to try to set up some docs and see how it goes before proposing to merge this.


ac-arcana:
Okay so for Read the Docs I've run into a snag. Normally it would automatically configure the repo to allow Read the Docs to function, but because I don't own the repo I don't have enough permissions to do that. It requires a web hook.

The manual instructions for setup are:
image

@Phazorknight, would you like to try setting this webhook up? Alternatively you might be able to set up Read the Docs yourself and it would auto configure. If this is all too inconvenient/too much work for what you feel are unneeded features then I would accept that answer as well.

Phazorknight:
>@Phazorknight, would you like to try setting this webhook up?

Ok, i just added the Webhook and it seems like the ping was successful.

ac-arcana:
Thank you for adding it. I'm still getting some errors form setup. Mainly that I don't have a .readthedocs.yaml set up, even though I do in this PR. Maybe it has to be in the main and can't just read from the PR, despite the ability for it to build docs from PRs...

Phazorknight:
Might be something that trips up ReadTheDocs on the initial setup. With your new permissions you should be able to push stuff to main now though, shouldn't you?

ac-arcana:
I can't on my own PRs.

image

image

Phazorknight (approved):
> I can't on my own PRs.

Oops, wasn't aware of that.
I just approved this PR now without proper review to enable possible push to main.

ac-arcana:
Okay so I was able to merge this. Almost there with getting this setup. I just forgot some quotes in the .readthedocs.yaml file.

PR #194, opened 2024-05-16
Added Grid Inventory that can be enabled or disabled

PeterD23:
- Added Grid bool so grids can be turned on or off at the designers
  convenience
- Hotbar is chosen by item order from top left to bottom right
- Quantity, Ammo and item textures scale based on the grid size of the
  item
- Green highlights if an object can be placed, yellow if it will swap
  out a colliding object, or red if it collides with multiple objects
  or doesn't fit
- Should work with controller as well, but not fully tested


Phazorknight:
This is mighty impressive! Great job! Especially cool that it can really be toggled on/off with relative ease.

- Bug: Items that are bigger in the inventory can "override" existing items. I had a health potion in the 3rd slot, then picked up the LaserRifle and it placed it self "on top" of it, basically deleting the health potion from my inventory.

- Bug: While the items themselves show up on the Hotbar, activating them doesn't seem to work quite as expected. Sometimes I was able to trigger an item with a different Hotbar key than what it is assigned in the UI.
To be fair, the Hotbar system should be revamped, so items can just directly get a button assigned instead of assuming the first 4 slots.

- Would love to see some tool tips in the exposed inspector settings, especially that when `grid` is enabled that the number of inventory slots needs to match the total of the inventory size for this to properly work as expected.
(eg. when the inventory size is 8x4, there needs to be 32 inventory slots in the inventory slot array).

PeterD23:
> This is mighty impressive! Great job! Especially cool that it can really be toggled on/off with relative ease.
> 
> * Bug: Items that are bigger in the inventory can "override" existing items. I had a health potion in the 3rd slot, then picked up the LaserRifle and it placed it self "on top" of it, basically deleting the health potion from my inventory.
> * Bug: While the items themselves show up on the Hotbar, activating them doesn't seem to work quite as expected. Sometimes I was able to trigger an item with a different Hotbar key than what it is assigned in the UI.
>   To be fair, the Hotbar system should be revamped, so items can just directly get a button assigned instead of assuming the first 4 slots.
> * Would love to see some tool tips in the exposed inspector settings, especially that when `grid` is enabled that the number of inventory slots needs to match the total of the inventory size for this to properly work as expected.
>   (eg. when the inventory size is 8x4, there needs to be 32 inventory slots in the inventory slot array).
Thanks for the feedback! I'll take a look and investigate the items getting overriden issue and the hotbar slots. I'll also look into adding tooltips, although the inventory slots array should automatically get resized to fit the X and Y dimensions specified. I was meaning to hide that from the inspector since the starter inventory array would be the grid-agnostic replacement.

Phazorknight:
The more I think about it, the more I feel like this could potentially be a permanent replacement of the "legacy" inventory system. I'd have to dive a bit deeper into the code and would like to give polishing the UI a shot, but here are my thoughts:

- If devs want to stick to the legacy system, they can by just keeping all items to 1x1 size.
- I wanted to rework the hot bar slots to work that they can be manually assigned, decoupling them from the main inventory resource.
- The only thing that's a bit trickier would be filling containers with items, would have to implement a check that would make sure the items defined in the inventory slots of a container don't overflow the container's defined inventory size.

Let me know if and when you'll work on certain parts on this, 'cause I might want to push an update to this PR myself next week. 😇

PeterD23:
> The more I think about it, the more I feel like this could potentially be a permanent replacement of the "legacy" inventory system. I'd have to dive a bit deeper into the code and would like to give polishing the UI a shot, but here are my thoughts:
> 
> * If devs want to stick to the legacy system, they can by just keeping all items to 1x1 size.
> * I wanted to rework the hot bar slots to work that they can be manually assigned, decoupling them from the main inventory resource.
> * The only thing that's a bit trickier would be filling containers with items, would have to implement a check that would make sure the items defined in the inventory slots of a container don't overflow the container's defined inventory size.
> 
> Let me know if and when you'll work on certain parts on this, 'cause I might want to push an update to this PR myself next week. 😇

Hey, I've fixed the issues you mentioned, details of what broke and how I fixed it are in the commit details. I'll be away to London next week so any other feedback I can look into after I get back!

ac-arcana:
I'm still working on getting Read The Docs to build, so for now please ignore any errors that the check has failed from Read the Docs for the purpose of merging.

Phazorknight:
Pushed some updates and will now try to merge this into main. 
Since I've made a few structural changes on main, this might take a few updates after the initial merge to fix what will inevitably break on the merge, so stay with me.

PR #200, opened 2024-05-25
Read the docs errors

ac-arcana:
So RTD is finally starting to build. It has an error with the theme. I've changed some settings which will hopefully address this error.
At this point there is a build in the repo that functions and includes content of both the readme and documentation.md. It is found at docs/_html/index.html


Phazorknight (approved):
Approving.
Thanks for setting this up. Will start reading up on how to fill it with info.

PR #224, opened 2024-07-11
Fix loading a existing scene state

mrezai:
Fixed by adding the scene name to file path.


Phazorknight (approved):
Looks good!
Thanks so much for this!

PR #225, opened 2024-07-11
Fix rifle firing, prevent it when reload animation is playing

mrezai:
and prevent player from shoot in the foot by excluding it in raycast.


Phazorknight (approved):
Nice, thanks for this!

PR #226, opened 2024-07-16
Fixes some player's physics issues.

mrezai:
Update stair stepping, now it's more compatible with Godot Jolt, previous code is a port of my Godot 3.5 implementation and new code added from updated version of main repository: https://github.com/mrezai/GodotStairs
New node added to player as body to handle rotation instead of using player's collision, useful for player with box collision(prevent player from unwanted fall on ledges).
Separation of computed velocity from move_and_slide's velocity, previous code cause unwanted damage to player for example after jump on objects.

### DIFF
--- a/COGITO/CogitoObjects/cogito_player.gd
+++ b/COGITO/CogitoObjects/cogito_player.gd
@@ -106,14 +106,17 @@ var stamina_attribute : CogitoAttribute = null
 var visibility_attribute : CogitoAttribute
 
 ## STAIR HANDLING STUFF
-var is_step : bool = false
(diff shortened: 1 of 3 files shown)


Phazorknight:
Woah, thanks so much for this. It's gonna take me a hot minute to fully check and test this.

mrezai:
Sorry I forgot to update save/loading of rotation in player state. I add it to current PR.

Phazorknight (approved):
Wasn't able to do a super-in-depth review, but in my tests, this all seems to work well.
Thanks again for this!

PR #229, opened 2024-07-17
Add camera smoothing

mrezai:
This PR adds camera smoothing to player.
PR sets as draft because by this implementation, animations in player's AnimationPlayer like jump doesn't work correctly and I'm not sure what is the best way to fix animations(tween, some formula in _physics_process ...).

Physics ticks per second = 30, monitor refresh rate = 120

Without camera smoothing:

https://github.com/user-attachments/assets/d63babd6-23ad-4eaa-ac0a-ae89b7241c11

With camera smoothing:

https://github.com/user-attachments/assets/24d4c252-998d-4ea3-a623-9076cccdcf31


mrezai:
Camera animations fixed by add a new node named CameraHolder as parent of Camera and move transform interpolation to it.

Phazorknight:
Hey, just getting around to checking this out.
I was wondering what case there is for reducing physics tics down to 30?

mrezai:
Hi, that config was only to show the effect better, physics tick hasn't changed in project.

Phazorknight:
Hey, just wanted to follow up on this:
I'm hesitant in merging this, as this currently isn't fixing any issues or adding major new features.
The smoothing could be useful, but as it's setting nodes tied to player motion to top_level, fearing this could cause unexpected behaviour with other things.

Another thing is that Godot 4.3 announced to add 3d physics interpolation soon-ish, so we might have a engine-native solution soon as well.

mrezai:
Hi, it's OK to close this PR to prevent complication but I think even with builtin physics interpolation, camera should interpolated using a similar method to achieve best result. Godot 3.5 has physics interpolation and this is related document:
https://docs.godotengine.org/en/3.5/tutorials/physics/interpolation/advanced_physics_interpolation.html#cameras

Phazorknight:
Closed due to migration to https://codeberg.org/Phazorknight/Cogito/pulls/229

( @mrezai sry for kinda abandoning this! 😇  If you want to do a quick update on Codeberg, I'm sure we could get this merged as it's probably still relevant! )

PR #233, opened 2024-07-19
Fix player rotation after scene transitions

mrezai:


Phazorknight (approved):
Nice, looks good!
Thanks for this!

PR #235, opened 2024-07-22
fix readme under development warning

yoshimatsudev:
fix under development warning misspell on Readme.md


Phazorknight (approved):
Thanks for catching that!

PR #236, opened 2024-07-24
Flashlight turn on/off while walking fix + more features

KrysisDK:
- Fixed issue where flashlight does not turn on / off while running, walking, and spam clicking.
- Added cooldown for flashlight, so it cannot be rapidly turned on / off (for better immersion, but can ofc be adjusted)
- Flashlight animation will finish before it will play again (better immersion!)


Phazorknight (approved):
Amazing. Thanks so much for this!

PR #240, opened 2024-08-02
Flashlight: Better Commenting + Restructuring

KrysisDK:
- Added much better commenting for future developers.
- Re-structured variables for easier reading.


Phazorknight (approved):
Thanks for this! Looks much cleaner.

PR #248, opened 2024-08-15
Fix FootstepSurfaceDetector raycast bug

mrezai:
Fix FootstepSurfaceDetector raycast bug by excluding its parent, now it's compatible with Godot Jolt.


Phazorknight (approved):
Looks good from my end! Thanks!

PR #253, opened 2024-08-22
Fall/Jump Landing Audio

niefia:
Basic attempt at a Materials based Fall landing audio player, using the same logic as the Step audio to determine materials.


Phazorknight:
hey, this is pretty cool!
Default volume seems to be quite loud, so finding a way to dampen that would be good.
I appreciate you having this separate to the footstep system, but I'd probably combine it. I don't see a case where we want to have these two as separate components.

niefia:
There was a few reasons I did it that way but in retrospect it may not be necessary.

I put together a quick test using the FootstepPlayer only, and came across one major problem which is:
If the player goes from landing into a run, the step sounds will overwrite the landing sound

Can't think of a way to avoid this aside from using 2 nodes for it, but if you have any ideas feel free to suggest. 
I'll work on cleaning up the code a bit in the meantime

niefia:
Ok code has been majorly cleaned up and with the last commit, it's been combined into a single AudioStreamPlayer3D node that avoids the sound overwrite problem I was having.

Phazorknight:
I would've suggested to maybe use the included QuickAudio plug in that spawns/pools Audio players as necessary.
But will take a look at your solution in a bit.

Thanks!

niefia:
Oh a plugin wasn't even needed, I just missed a basic Godot feature. Just needed to set the AudioStreamPlayer3D to have polyphony to fix it, so set that to 8 max now instead of 1.

Phazorknight (approved):
Nice job cleaning this up!
Let me know if you're still planning on working more on this PR or if it's ready to merge.

niefia:
It's ready to merge now. Functionality seems to be perfect from testing, and just put out a last commit to improve the readability of variables and comments before it gets merged.

Phazorknight:
Amazing work, thanks so much for this! Merging.

PR #256, opened 2024-08-28
Fixed fridge inventory update

antono2:
The laptop quest requires an inventory update on the kitchen_fridge to succeed.


Phazorknight (approved):
Ah, thank you! I didn't even realize that this quest broke again.

Appreciated!

PR #258, opened 2024-09-02
Fixed player detection for Pressure Plates

niefia:
Fixed player detection on Pressure Plates by adjusting some parameters. Previous Lower height seems to cause player to never be recognised. Previous Lower Mass & Higher gravity scale seem to cause a bouncing effect with player.


Phazorknight (approved):
Neat! Thanks so much for this!

PR #261, opened 2024-09-04
Buttons & Switches interact to all hits

niefia:
Buttons previously only activated with Hitscan weapon, now should activate with Projectiles.
 Switches now react to both Hitscan & Projectiles.

Technically also possible to hit with Pickaxe but difficult to line up due to Pickaxe collision box being tiny and not intersecting with the players crosshair. May be worth enlarging Pickaxe Collision area


Phazorknight:
Hey, thanks so much for this! Looks pretty good, but I'd reckon this behavior should be easy to switch on/off (and maybe even more fine-tuneable?

There's games where I wouldn't want all switches to be triggered by shooting at them. Also _PREY 2017_ comes to mind, where only the foam darts let you activate the display buttons...

niefia:
We may have to edit the Damage system then to tell us what caused the damage as currently it'd be impossible to do a setup where Soft Projectiles can work and Hitscan damage can't.

Hitscan = true,  Projectile = true - Current setup, activates on any damage
Hitscan = true, Projectile = false - Possible by disabling CogitoProperties for the Switch, thereby blocking Soft projectile damage
Hitscan = false, Projectile = true - Currently impossible as there's no way to distinguish hitscan from projectile damage
Hitscan = false, Projectile = false - Possible by disabling Health for the Switch

I started writing a hacky workaround that passes item reference through from the weapon to the func damage of the hitbox, but that seems against the immersive sim spirit since we should probably rely on some sort of "damage type" rather than using specific references

Phazorknight:
Yeah, I'd hold off on that until the whole `cogito_properties` system is more robust and fleshed out.

I don't see this as a high-prio feature.
Thinking out loud here, just adding a hitbox component to a switch and hooking up it's signal to trigger the switch should be good enough to make this feature work without having to actively make the switch itself more complex.

Then any future `cogito_properties` behavior could be part of the hitbox component.

niefia:
Just removed the health attribute to simplify it. 

Currently switches don't react to hits but buttons do, so it may still be worth merging this one just for the sake of having consistent behaviour between them.  With the idea of revisiting the issue in future once the damage system has been updated

Could also remove the CogitoProperties if you'd prefer that for now, but it'd cause no reaction from projectiles

Phazorknight (approved):
Finally getting around to reviewing this.
Appreciate your work to getting these nodes updated and consistent.

PR #262, opened 2024-09-06
Added Footstep system to NPC

niefia:
Added Footstep system to NPC by adapting the logic of the player footstep system.


Phazorknight (approved):
Nice implementation, thank you!

PR #265, opened 2024-09-10
Cogito Sittable

niefia:
Addition of the Cogito Sittable as proposed here #255 

Still WIP with the following issues needing to be handled before implementation:

- Look marker defining player horizontal sitting vision
- Sittable interaction behaviour on non standard player states; crouching, jumping, on ladders etc
- Controller support
- Sittable area Player detection moved back from process to signal


niefia:
@ac-arcana Continuing from the discussion in #255, I've tested the ladder interaction and doesn't seem to have any weirdness like that currently since on sitting, physics is disabled and the player just tweens to the sitting marker location. 

If you actively wanted to replicate that bug effect you'd probably have to do the movement through physics instead. 
Is this a common enough thing to introduce an alternate sit mode?

niefia:
Wanted to test some weird uses so made a Swing by just attaching a desk chair to hinge node and here's the results:

https://github.com/user-attachments/assets/f22b6332-3bea-4b9d-a3c9-22f4e6dcec09

Don't think its too bad given I hadn't even consider this case at all, but I can see what needs changing to work better

Basically just need a dynamic tween if its a physics object in case it moves mid tween, also the vertical look parameters need to be defined by the Sittables axes.

For performance reasons it might also be worth having a Static or Physics option for Sittables so that Static ones skip all the extra things required for Physics sittables.

niefia:
Thanks for the feedback, the seat to seat movement was an early consideration as the sofas being wide enough for 2 Sittables made me think it was needed.

Working on implementing different options for the on leave behaviour, with the following types: 

	ORIGINAL,  ## Player returns to original location - Current Method
	AUTO,    ## Player sent to random nearby available location 
	NODE     ## Player is placed at defined Leave node


Also saw the discussion on renaming the Sittable to something more broad, but don't want to change it until a consensus is reached on what to call that really.

 My thought on name is that `CogitoSittable` is probably fine for it since it covers most uses and we'll aim to demonstrate the less usual ones for that. If we're ok with a Lift belonging to the `CogitoDoor` class, instead of something more generic like "Mover" then I think we should just keep the Sittable name.

 I think Sittable is also more immediately understandable than something like `PlayerRestrainer`, despite it being less accurate, it is more approachable

Phazorknight:
@niefia _re naming:_ I see your points and agree. Let's keep the naming _Sittable_

In regards to "leave behavior": I might rename the _Node_ option to something like _Transform_ ? Unless there's other requirements needed/used, but Transform makes it clear as to what is being referenced.

How would the _Auto_ behavior work?

niefia:
The idea for the Auto method was to use Navmesh. Give the player a Navigation agent and test random nearby locations until a valid location is found. Users can define the parameters of the testing, and if it gives no results after an amount tries it will fall back to another method.

My current attempt _technically_ works, but barely. It needs way too many tries, and can send player through the floor (despite not moving player in Y). If it just proves too unreliable I can either cut it, or just leave it to implement in future

niefia:
Managed to implement the Auto method now, currently it only works on the Laboratory map since that's the only one with a Navmesh.  

Weirdly allowing the y-axis some degree of randomness actually helps it be more reliable in finding available locations even though it should be immediately overwritten by the navmesh_offset_y before it's even checked.

Phazorknight:
Any specifics of whats causing you trouble? Or something I should take a look at?

I know the save system isn't the best documented one (as it also just recently changed a bit), but happy to help or give pointers.

niefia:
Think I'm just a bit stuck with it since there's so many different errors It's hard to tell what's going wrong exactly.

May be more worthwhile for you to give it a go at this point, then I can focus on other bits of polish in the meantime.

Phazorknight:
> Think I'm just a bit stuck with it since there's so many different errors It's hard to tell what's going wrong exactly.

Took a bit of a look and I found one issue with the swing: 
`CogitoSittable` is a type of `cogito_object`, but you had it as a child of another `cogito_object`, which trips up the save system (cogito_objects should only have interaction components as children).

If you remove the cogito_object script from the swing and instead make the impact sounds a component of the Cogito_Sittable, it should work (did a quick test and it seemed to save/load as expected).

image


The default cogito_object is part of the `persistency`group (which means, it gets destroyed/respawned on loading a saved game). 
Most sittable wouldn't be that, and you correctly instead added it to the `save_object_state` group instead.

niefia:
Nice job spotting that, it was just a leftover from the Swing being based on the Desk chair. 
Weirdly the desk chairs technically work, despite being setup in that same broken way. I'll reorganise those too. 

Experimenting with this has taught me some more things about the save system, and that most of my problems are coming from the physics Sittables, so I'll try and just tackle functionality for static Sittables first.

niefia:
Turns out that was basically the cause of all of the issues I was having, just through a bunch of different errors.

So fixing that allowed me to pretty easily implement the rest of the code I already had, and it all seems to work fine now. 
I just got stuck because my save code was technically all correct, it's just that node structure broke everything.

Still need to fix up a few more things before its ready to merge, but thanks for much for finding that

niefia:
All the major issues have been sorted now. Including a bug not even related to Sittables

The only thing that remains is an issue with the Swing, due to the hinge not currently being saved. Trying to sort it without giving the hinge its own script.

Phazorknight:
> All the major issues have been sorted now. Including a bug not even related to Sittables

awesome, let me know if its ready to merge.

> The only thing that remains is an issue with the Swing, due to the hinge not currently being saved. Trying to sort it without giving the hinge its own script.

what exactly are you trying to save?

niefia:
https://github.com/user-attachments/assets/8e12aa2c-dd29-45be-9f8d-ff438886822b

This is the issue mentioned above, the Swing state is currently not being saved. It's the parent of the Sittable so can have weird behaviour if there's a mismatch. So need a good way of saving that too.

Another issue is this:


https://github.com/user-attachments/assets/bbae66a5-433e-4fd6-a294-9f1cfe6ea393

Sometimes it loads in the correct position, sometimes not. Sometimes it loads correctly once, then never after. Occasionally goes back to its original location too.

Seems like there's a lot more to do, so will have another attempt at fixing these tomorrow. The save system is very broken for physics sittables currently.

Phazorknight:
> Sometimes it loads in the correct position, sometimes not. Sometimes it loads correctly once, then never after. Occasionally goes back to its original location too.

The sittable vehicle is its own class, right? I haven't looked into that one that much yet, but shouldn't be too hard to fix, I have a hunch of what's wrong.

> Swing state is currently not being saved
I assume you mean the position/angle of the swing? Honestly, if that doesn't get saved but everything else works, then I'd be fine with that.
Alterantively, we could simply add an optional reference to it to the sittable and then save any value we'd need to pass to to it as part of the sittable saved values.

niefia:
yeah vehicle is its own class

The problem with the swing is a bit deeper than that. If the player is sitting on it when the swing position & angle don't get saved, but the Sittable does. 
It can cause an issue where the player is placed in the saved location relative to the swing in the new location.


https://github.com/user-attachments/assets/33f77535-afbe-481d-b38b-d2f4be5cc8ce

> Alterantively, we could simply add an optional reference to it to the sittable and then save any value we'd need to pass to to it as part of the sittable saved values. It has some limits, but might be "good enough" without overcomplicating things.

This is what I was considering too. But need to consider making it very general and not specific to the scenario of a swing.
Are there other instances where we may want to package multiple nodes together under the save of one?



Realised there's also an issue that has been introduced with the previous commit where the seat interaction doesn't appear on load, so need to sort that too.

Phazorknight:
**RE vehicle:**
So I took a quick look: I saw that the `cogito_vehicle` inherits from `cogito_sittable`. The sittable is using the "save_object_state" group and it's behavior, which wasn't really designed for RigidBody objects like the chair (which originally was a `cogito_object`).
My suggestion would be to separate `cogito_vehicle` out from the sittable, instead maybe base it on `cogito_object` or even just have it's fully it's own.

instead of:
`add_to_group("save_object_state")`

it should be:
`add_to_group("Persist")`

which should remedy the issue of loading/saving the position correctly. This of course could cause new errors if the player is loading a game where they're sitting in the vehicle. One way would be to add a method to re-seat the player on load, that also includes some basic checks that could "eject" the player into a vaild unseated position to avoid glitches like the player sitting in a chair that has been flipped upside down or in a wall etc.

niefia:
Would have to apply this to the whole Sittable class then since vehicles aren't the only physics body, there's also physics Sittables that will need the same data stored. All the Vehicles add is the player input part really, not the physics.

I do have a check and optional eject for player if the chair falls so probably able to integrate that at least.

I guess the Persist group is going to have to be the solution but it does seem like it will cause a bunch of issues with the Sittable being deleted and recreated on load.  Like will the reference to the current active Sittable still be valid on node recreation? 
Just annoying that Sittables really don't fit with either of these groups really.

Phazorknight:
> Would have to apply this to the whole Sittable class then since vehicles aren't the only physics body, there's also physics Sittables that will need the same data stored. All the Vehicles add is the player input part really, not the physics.
> I guess the Persist group is going to have to be the solution but it does seem like it will cause a bunch of issues with the Sittable being deleted and recreated on load. Like will the reference to the current active Sittable still be valid on node recreation? Just annoying that Sittables really don't fit with either of these groups really.

The main purpose of the `Persist` group is for items that the player can put in their inventory, drop from their inventory or destroy otherwise. Not to say we can't make destructable sittables or vehicles, but I'm saying that we can do that without having to add them to this group.

Personally, I think this has reached one of these points where putting in a limitation makes the most sense in order to keep usability and complexity. I'd keep `sittables` as `save_object_state` objects to avoid any spawning issues.

And I'd also not try to make the `vehicle` class compatible with RigidBody objects, this problem basically just occured cause of how the office chairs are set up as one of those. I'd be fine with basically forcing COGITO users to choose:
Ok, is this a classic rigidbody object that's gonna be carried around, knocked over, destroyed etc? Or a vehicle (probably based on AnimatableBody or similar) which will _not_ have the same behavior as a RigidBody but in exchange offer other (more fitting) features?"

niefia:
May have an idea on how we can keep Rigidbody3D functionality and have Sittables & Vehicles in the `save_object_state` group. 

So the issue is coming from using position to set physics bodies which gets overwritten by the physics server, that's what's causing the inconsistent behaviour.

If we apply a physics transform on load to put the Sittable in It's correct location. It should be possible.

Have a very early implementation of this and it does seem to work.

niefia:
It may have been even simpler than that, just needed global position instead of local position in saves.

Which I did think was the cause earlier on but had really weird effects when I tried to use it, and didn't seem to even work correctly for the nodes it should have. 

But anyway, It's done now.  Sittables are now working with the save system as intended without compromising on functionality. 

Just need to finish some things up and run some more tests, so probably going to be a bit until its finally merged

niefia:
Fixed up all the bugs that I'm aware of, and added a whole bunch of features to improve the stability of things.

Would be great if somebody else is able to properly test it out before Merge, as I wouldn't want to merge such a large PR without further tests.

Phazorknight:
Gave it a quick look and run into two thing:
1. Added the Swing to the Laboratory demo scene and noticed it didn't work right out of the box. Needed to activate "physics sittable" to it. As this is it's own packed scene, I assume it's fine to keep that.

2. I saved my game while sitting on the swing. Then got up and attempted to load that save.
Errored out with:
`Invalid set index 'sleeping' (on base: 'StaticBody3D (CogitoSittable)') with value of type 'bool'.`
from CogitoSceneManager at function `load_scene_state` line 302: `node_to_set.sleeping = true`

niefia:
So I had removed the Swing, since it had a lot of unique issues that would only ever be present for a Swing. Intended to remove it entirely but hadn't got round to it.

However have had another try and managed to get most things working for it. Even the things I thought wouldn't.
Can't guarantee its 100% perfect though

Phazorknight:
Did some tests and ran into a few issues that were a bit hard to reproduce (mostly on level scene changes, one time I had all rigidbody sittables double spawn in the lobby, but couldn't reproduce, might've been a scene state file issue).

One bigger issue, was able to get myself "trapped" in a sittable, here are the steps:
- Start Lobby scene.
- Sit down at the first blue chair on the left, facing entrance.
- Save game.
- Get up and start walking/sprinting towards conference room/archive room.
- Load game.
- Player is seated in blue chair but faces towards the conference room and is thus unable to aim at the "Stand up" interaction, staying trapped.
- Also noticed that if I was sprinting when I loaded, my stamina kept going down. This could've been related with the cogito_player.gd issues introduced with #283 though.

One suggestion in regards to saving and loading on a sittable: I'd probably just "eject" the player to the "leave" node (or general leave position depending on the _placement on leave_ selection. That might be easier to implement.

I've been playing a bit of _Abiotic Factor_ that also has placeable and movable objects the player can ist on and AFAIK that game doesn't save/load the player's seated state.

niefia:
Not able to replicate the first issue, so can't really comment on that one. If we can get it happening again will investigate further.

For the other issues I was able to solve them all, so looking pretty good to merge now unless you can find any more bugs.

Oh and the sprinting issue is definitely one introduced in #283 since I've encountered it on other branches. And think I first noticed it with the fix I did after that Commit.


Should be fine to keep save/load chair state, since it seems to work fine now.
I think I probably got the idea of Office chair vehicles from Abiotic Factor funnily enough.

niefia:
Just checked now and yeah quite a few issues with sprinting were introduced with the [Trying to stop player from standing up while opening a GUI](https://github.com/Phazorknight/Cogito/commit/71fcf4c14104136af527a1a961c5a8ea2a4b1994) commit. All seems fine again if you revert back

Phazorknight:
Wow, this is getting really robust. Great work!
Saving and loading all seemed to work.

Ran into two issues:
1. The swing still seems to have a `Carriable` component. Was able to "break" it by starting to carry it, and suddenty the swing froze in place and I couldn't interact with it anymore either (neither carrying nor sitting. I assume removing the Carryable component will fix this.

2. When saving while on the swing, then loading again, the mouse look seems to be limited to only up/down. Once I get up and sit down again, the mouse look seems to work again with left/right as well. I'd say this is very very minor.

Seriously fantastic work on this! Thank you again!

niefia:
Removed the Swings carriable component and fixed up a few other things

Can't seem to replicate the second issue now, but think I may have encountered it previously. If it happens you can usually regain look by moving mouse side to side rapidly.

Phazorknight:
> Can't seem to replicate the second issue now, but think I may have encountered it previously. If it happens you can usually regain look by moving mouse side to side rapidly.

Yeah, trying this repeatedly now and it seems to happen only at specific moments while swinging in the swing, so it might be tied to the physics tics and/or the player controller in general. Anyway, like I said, not a big issue and overall this works really well now.

I'm up to merge this into main if you say it's ready.

niefia:
It's ready to merge, It's had a week of bug testing and fixes, and plays very stable.

Thanks for the help with bug testing

Phazorknight (approved):
Approving to merge!
Again, thank you so much for this.

PR #270, opened 2024-09-13
NPC Door handling

niefia:
really simple door handling i had done, thought i should submit since it was relevant to #269


Phazorknight (approved):
Nice work! Thank you so much.

PR #275, opened 2024-09-17
Fixed backwards text on Platform

niefia:
Lift platform/Lower platform text was opposite of what it should be, so just swapped them around to fix


Phazorknight (approved):
Oops, thanks for catching this!

PR #281, opened 2024-09-20
Interaction Ray + Shapecasting for easier object interaction

OvercastInteractive:
- Uses the existing raycasting behavior for when aiming directly at the target interactable
- If not aiming directly at the target interactable, a shapecast is used to detect any nearby interactables
- If raycasting a body that is not interactable, such as a tabletop, moves a 'hotspot' node to the raycasted point
- The hotspot point is used as a reference for any shapecast hit colliders to compare distances with all other hits
- The closest interactable collider to the hotspot is selected as the target interactable


Phazorknight:
Hey,thanks so much for this, this is quite simple but yields amazing results!

I get a weird issue though:
In the Lobby level scene, everything works as expected, but in the Laboratory scene, the `interaction_raycast.gd` throws an error. Seems like the `hotspot` and `highlighter` reference are all null, which baffles me, as they're set `onready`

This happens whether I start directly in the Laboratory scene or transition to it from the Lobby scene.
I've tried to copy the Player.tscn from the Lobby to the Lab to see if there are any weird overrides, but the error persists. ☹️

OvercastInteractive:
I expected the highlighters to be removed after a little more testing (as I found looking into cabinets and drawers toward multiple carryables produced unexpected/undesired detection results) and the hotspot can be converted into a `Vector3` or instantiated on `_ready()`. But you're right, when I enter my own scenes from the Lobby I don't have this issue. Only when going into the Laboratory. In the Laboratory, when toggling editable children to true they appear nested under the InteractionRaycast node, as expected.

OvercastInteractive:
In order to get the Laboratory scene to work I had to add null checks and assignments for every null reference at the top of `_update_interactable()` and also check if each was null before updating their visibility and position, like this:

if hotspot == null:
	hotspot = $ShapecastHotspot

if hotspot != null:
	hotspot.transform.origin = to_local(hotspot_global_position)


It's like their references are lost or cleared before every call. 🤨

OvercastInteractive:
The issues are due to the Enemy. I didn't realize they were using the same code, sorry about that! Here's the error code: `E 0:00:19:0588   interaction_raycast.gd:8 @ _ready(): Node not found: "InteractionShapecast" (relative to "/root/Cogito05DemoLaboratory/CogitoEnemy/InteractionRaycast").`

Phazorknight:
Aaah, the enemy is pretty new so I totally forgot about this!
I've just pushed an update that should fix all of this (just see the description).

Take it for a spin, but I'd wager this is now ready to merge.

OvercastInteractive:
I made a couple of minor changes, but yeah, everything now seems to work well!

Phazorknight:
Awesome! Again, thanks so much for your work on this. You implemented this way better than I could have.

PR #282, opened 2024-09-21
Fix - Audio volume change update signal

niefia:
Previously Audio volume only changed on restart, as update signalling was not properly connected

Fixed that here, so now Volume can be changed at any time


Phazorknight (approved):
Nice catch, thanks for this!

PR #284, opened 2024-09-22
Added Physics save/load to RigidBody CogitoObjects

niefia:
Added save/load for the Physics properties of RigidBody3D CogitoObjects

Thrown CogitoObjects will now continue their trajectories on save/load instead of just falling directly down from the point they were saved at.


Phazorknight (approved):
Niiiiice! Thank you!

PR #287, opened 2024-09-23
Remove & gitignore .tmp files

niefia:
Noticed a temp file was added in a recent commit, and realised there were a few in the project.
So for Clean-up removed them and added .tmp file type to the .gitignore to prevent them being added in future

Seems to stem from a bug where the temp files are left: https://github.com/godotengine/godot/issues/82270


Phazorknight (approved):
Thanks for this! I hadn't even really noticed it but good to get rid of.

PR #289, opened 2024-09-24
Spawn on Death Array + NPC corpse

niefia:
Changed Health attribute to have a Spawn on death array, to allow multiple scenes to spawn on death.
Used this functionality in adding NPC corpse, on NPC death.


Phazorknight (approved):
Using an array makes sense.
Also a smart way on how to add the "corpse"!

Thanks for this!

PR #290, opened 2024-09-24
Fixes for Currency update

niefia:
Know the currency update is still WIP, but just trying to patch out an error that breaks saving/loading due to no set_currency function in the currency.

Also replaced coins collision shape it it was colliding weirdly with surface, seems fine with a box


Phazorknight (approved):
Thanks so much for catching this!
I thought I had added this but apparently not. 😅

PR #291, opened 2024-09-25
Added projectile stick_on_impact

niefia:
Projectile stick_on_impact property, could be useful for implementing arrows, crossbow bolts, rope arrows etc

Haven't enabled as this likely shouldn't be default behaviour but useful as an option

Had a more complex version that reparented the projectile to the collision object, but found using pin joint seemed more reliable and versatile


Phazorknight (approved):
Awesome, love it!
Thanks so much.

PR #292, opened 2024-09-26
Lock Interaction Component & Dual Interaction Rework

niefia:
Reworked the door locking to match the proposed ideas in #269

Introduced Lock Interaction Component to support this


Phazorknight:
Oooh, making locks their own interaction component is not a bad idea, opens some cool possibilities actually.

Could you include a component scene for the lock interaction as well?

Personally, I'm not the biggest fan of needing two different interaction buttons to unlock then open a door, but using the DualInteraction component, this can be remedied (press for open/close, hold for lock/unlock). 
Will play around a bit with this a bit to see how this integrates.

niefia:
Still need to solve a couple of problems before merging this one:

First is when to run the key check, as running it on the end of hold down feels wrong, like it feels like you should have been using the key during the hold down time. May have to redo key check and run it at the start of an interaction to see if its valid. 
Also means adding the functionality of preventing the hold interaction if the key is not found.

Second is that the interaction text label doesn't seem to update until the player looks at the object again.  Can't seem to figure out what's causing that one.

niefia:
Managed to solve both issues, also updated the title since the PR has also become a Dual interaction component rework at this point.

Give it a try and let me know what you think

Phazorknight:
Great work! The way this is set up makes sense and fits.
Also reminds me that I wanted to update the `Hold UI` to ideally be like the background of the interaction prompt or something similar....

Phazorknight (approved):
Should be good to merge in! Thanks so much for this!

Issue #250, opened 2024-08-17
Cogito crashes on editor startup


YarasaSpacekid:
Hello! I'm currently using the most recent download of cogito (both from github & the godot project manager versions), on godot version 4.2.2-- cogito seems to crash on startup :'> My initial theory was that godot's forward+ mode was the issue (as it causes **everything** to crash on this computer), but it seems that even gl compatibility mode has the same issue.

Currently running on a linux computer, with sorta-kinda-almost 8GB of ram

YarasaSpacekid:
Running the game (instead of editing) makes a strange "empty" game with buttons that don't do anything. Because I can open, run, and close this version without crashing, I can actually get a text log of the terminal output :0 I'll paste it here.

```
Godot Engine v4.2.2.stable.official.15073afe3 - https://godotengine.org
Inconsistent value (1) for DRI_PRIME. Should be < 1 (GPU devices count). Using: 0
OpenGL API 4.5 (Core Profile) Mesa 24.0.9-0ubuntu0.1 - Compatibility - Using Device: Mesa - AMD SUMO (DRM 2.50.0 / 6.8.0-40-generic, LLVM 17.0.6)
 
Running project: /home/[USER_DIRECTORY]/Games/godot/Not By Me/Cogito-main
Godot Engine v4.2.2.stable.official.15073afe3 - https://godotengine.org
Vulkan API 1.3.274 - Forward+ - Using Vulkan Device #0: Unknown - llvmpipe (LLVM 17.0.6, 128 

SCRIPT ERROR: Parse Error: Identifier "OptionsConstants" not declared in the current scope.
          at: GDScript::reload (res://COGITO/EasyMenus/Scripts/startup_loader.gd:9)
SCRIPT ERROR: Parse Error: Identifier "OptionsConstants" not declared in the current scope.
          at: GDScript::reload (res://COGITO/EasyMenus/Scripts/startup_loader.gd:14)
SCRIPT ERROR: Parse Error: Identifier "OptionsConstants" not declared in the current scope.
          at: GDScript::reload (res://COGITO/EasyMenus/Scripts/startup_loader.gd:15)
SCRIPT ERROR: Parse Error: Identifier "OptionsConstants" not declared in the current scope.
          at: GDScript::reload (res://COGITO/EasyMenus/Scripts/startup_loader.gd:16)
SCRIPT ERROR: Parse Error: Identifier "OptionsConstants" not declared in the current scope.
[...]

PR #293, opened 2024-09-26
Attempted Weird Linux Bug-Fix

ZeroHanekoma:
I'm not sure if this is related to #250 but my issue is very similar.  This is not yet ready for merge because there are still issues with the wieldables, and likely other things I haven't noticed yet.
Goal here is to figure out the issue with some Linux systems being unable to load the latest build of Cogito.  Using 4.3 but should be no difference in versions, afaik.

What I did:
1. Re-forked the current Cogito repo
2. Cloned that and then deleted almost everything
3. copied over an older build of Cogito I had
4. Copied over only the new code, leaving Assets empty
5. Copied over godot.project to see if issues arose
6. Copied over almost every asset

I'm going to try copying over assets I haven't replaced yet to further narrow down the potential issue, but my only issue currently is the Wieldables don't seem to know how to reference their animations anymore.  If you can look at that, I'm very confused.  Anyone else on Linux who can't load the project, try this branch and see if you can load into it.

From what I'm finding, it's something in the Assets folder crashing, specifically a model but I can't figure out what.


niefia:
Wieldable animations are a known issue with Cogito in Godot 4.3, reported here https://github.com/Phazorknight/Cogito/discussions/247

Cogito hasn't been updated to 4.3 yet, so I'd suggest using 4.2.2 for now to help narrow down the issue your encountering

ZeroHanekoma:
I've tried it with 4.2.1, 4.2.2 and 4.2.3.  I almost thought it was a system issue until I started playing around with it more.  Unfortunately my logs and running it with --verbose doesn't tell me anything I don't already know.
Hoping some other Linux users can play around with it in here.

Thanks for pointing me to that issue I probably wouldn't have noticed it and continued trying to decipher what I did wrong on my own, lol.  I'll look into that on my own time then and double check it in Godot 4.2 when I wake up.

Phazorknight:
So Cogito works just fine on my linux laptop:
Godot Engine 4.2.1 stable
Fedora Linux 40

I would definitely advice against using Cogito with Godot 4.3 yet as it has not been tested or ported for it yet at all.

ZeroHanekoma:
Yeah and showing from #201 and from my own personal experience, that's not every Linux machine.  Something is borked somewhere for some users and that's what I wanted to figure out.
But alright man.  I'll just go back to another project and wait.

Phazorknight:
@ZeroHanekoma 
> that's not every Linux machine. Something is borked somewhere for some users and that's what I wanted to figure out.

For sure, and I'm curious what the issue exactly is. I provided of what I know on how Cogito runs on my end to narrow down sources of errors. The first step would always be to confirm the version of Cogito and Godot.

If the issue is with a specific asset, then making a new blank Godot project and importing just the Assets from the Cogito project might provide further insight.

PR #295, opened 2024-09-28
Lightmeter options + improvements

niefia:
For performance reasons I think Lightmeter settings should be accessible so added some options

Also Lightmeter camera now faces floor instead of facing forwards, seems better this way from my tests but if we'd prefer to keep it facing forwards I can change it back


Phazorknight (approved):
Makes sense! 
Also good to have these parameters exported.

Thanks for this!

Issue #283, opened 2024-09-22
Player exits crouching when opening Inventory or interaction GUI


OvercastInteractive:
**Cogito and Godot Engine Version:**
Godot version: 4.2.2
Cogito version: beta 202409.03

**Description:**
Interacting with a keypad while crouching locks you into the keypad UI (you can still exit it by pressing Esc)

**Reproduction steps:**
1. Go into a crouch
2. Interact with the archive keypad (it looks like you are immediately forced out of your crouch)
3. Attempt to exit the keypad with the interact button

**Expected behavior:**
The interact button works the same while not crouching, as a toggle that can enter and exit the keypad UI

Phazorknight:
Thanks for flagging this. Looking into this a bit more, it seems this has nothing to do with the interaction.

If you are crouched when interacting -> opening a gui like the keypad, the player character stands up and thus the interactable object is no longer being hit by the interaction_raycast due to the position change.
The interaction is not there anymore because the interaction raycast isn't hitting the object anymore.

That being said, I'd expect the behavior to be that the player character simply stays crouched when opening a GUI (like the keypad, readable etc.), so I'm looking into fixing that as part of the `cogito_player.gd` script.
So far it seems more complicated that I thought, unfortunately.

Phazorknight:
I'll probably have to undo some of the changes I applied to `cogito_player.gd` cause it seems like I've introduced some issues with the sprinting. Player doesn't stop sprinting when the sprint button is being let go, making it feel like they're floating a bit.

Phazorknight:
Renaming this issue to better describe whats going on.

niefia:
Figured the cause and and a quick fix for this:

**Cause** is this line (**554**) not being able to run `head.position.y = lerp(head.position.y, CROUCHING_DEPTH, delta * LERP_SPEED)`
So because movement is being paused, its causing that line to no longer apply, which is why the player is slowly moving back up.
It's nothing to do with the standing up code which is why it was hard to track down, its just the crouch code not being run.

For a quick fix can remove `!is_movement_paused` from this line (**542**): `if crouched_jump or (not jumped_from_slide and is_on_floor() and try_crouch and !is_movement_paused or crouch_raycast.is_colliding()):`

But not sure what other issues that might cause since the player code is a bit complicated. May want to refactor the player into a state machine at some point, especially since there's even more behaviours to add in future.

Phazorknight:
[...]

PR #299, opened 2024-09-30
Quick Fix for issue #283

niefia:
Quick Fix for issue #283. Explanation posted there 

> Figured the cause and and a quick fix for this:
> 
> Cause is this line (554) not being able to run head.position.y = lerp(head.position.y, CROUCHING_DEPTH, delta * LERP_SPEED)
> So because movement is being paused, it's causing that line to no longer apply, which is why the player is slowly moving back up.
> It's nothing to do with the standing up code which is why it was hard to track down, its just the crouch code not being run.
> 
> For a quick fix can remove !is_movement_paused from this line (542): if crouched_jump or (not jumped_from_slide and is_on_floor() and try_crouch and !is_movement_paused or crouch_raycast.is_colliding()):

### DIFF
--- a/COGITO/CogitoObjects/cogito_player.gd
+++ b/COGITO/CogitoObjects/cogito_player.gd
@@ -539,7 +539,7 @@ func _physics_process(delta):
 		try_crouch = Input.is_action_pressed("crouch")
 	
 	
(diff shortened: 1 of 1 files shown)


Phazorknight (approved):
Thanks for finding this! Yeah this behavior was hard to pin down.

PR #301, opened 2024-10-01
Prevent sound_on_damage_taken playing on Death

niefia:
Prevent sound_on_damage_taken playing on Death as this overlaps with the death sound currently.

Added a new field of sound_on_hit that will play on every hit including death as I didn't want to remove that functionality as an option. My idea with this is sound_on_hit could be used for something like a bullet hit sound. Whereas sound_on_damage_taken can be used for NPC response to damage


Phazorknight (approved):
Makes sense! Thanks for this!

PR #304, opened 2024-10-03
Sittable controller support

niefia:
Sittable controller support, only basic version that allows free look currently.

Refactored Sittables Process into it's own function _process_on_sittable


Phazorknight:
Hey, thanks for this. I've had a family emergency, so for the next couple weeks, I won't be able to work on Cogito as much as I used to. I'll trust all Cogito collaborators to review and approve any PRs and merges.

Phazorknight (approved):
Looks good from my end! Maybe not perfect but definitely improves the behavior.
Thanks for this!

PR #306, opened 2024-10-03
Modified security camera to use player light level for detection

niefia:
Modified camera to use player light level for detection

Renamed spot_time to detection_threshold to reflect this change

### DIFF
--- a/COGITO/CogitoObjects/cogito_security_camera.gd
+++ b/COGITO/CogitoObjects/cogito_security_camera.gd
@@ -8,19 +8,33 @@ signal object_no_longer_detected
 signal turned_off
 
 @onready var audio_stream_player_3d: AudioStreamPlayer3D = $AudioStreamPlayer3D
-@onready var lightmeter:CogitoAttribute = (CogitoSceneManager._current_player_node.player_attributes.get("lightmeter"))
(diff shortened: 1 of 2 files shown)


Phazorknight:
Hey, thanks for this. I've had a family emergency, so for the next couple weeks, I won't be able to work on Cogito as much as I used to. I'll trust all Cogito collaborators to review and approve any PRs and merges.

Phazorknight (changes requested):
A few thoughts on this:

- This now seems very dependent on the Cogito Player having a lightmeter component. I personally don't use that component at all, so I'd want this to be integrated as purely optional.
- Also worth keeping in mind if we'd want the camera to be able to detect other entities (like other NPCs or enemies)

Phazorknight (approved):
Nice, thanks for these updates.
I think that'll do nicely.

Let me know if you still wanna add to it or if it's good to merge.

niefia:
In future I'd probably want to revisit the non_player_detection_list, and make it use the custom class names instead of just names. 
But there's not a good way of doing that in the current version of Godot. 
Once we've updated to 4.3 might revisit since there is a way of doing that in 4.3: https://github.com/godotengine/godot/issues/21789

Other than that I'm happy to merge

PR #309, opened 2024-10-11
Fix Lobby overwrite

niefia:
Lobby changes from Sittable update had been overwritten by the Lock PR

This reverts to previous Sittable lobby version whilst re-adding neccesary Lock changes


Phazorknight (approved):
Checked and seemed all good on my end!
Thanks!

PR #311, opened 2024-10-11
Plug-in restructuring + COGITO Editor tab

Phazorknight:
- Moved all `/COGITO/` files to `addons/cogito/`
- Added plugin in script (so Cogito can be activated/deactivated in the plugin manager and it adds the autoload scripts as well.
- Added `Cogito` editor tab to offer some global options.


Phazorknight:
Little heads up: As this moves pretty much all cogito files to a different directory, some references might break (though I tried to update all of them). Might work better if dowloaded and tested in a blank project.

niefia:
Have only had a quick look at this so far, will review it properly tomorrow. But just a quick issues fix
Main menu points to incorrect scenes i.e `start_game_scene = "res://addons/cogito_plugin/DemoScenes/COGITO_3_Lobby.tscn"` instead of `res://addons/cogito/DemoScenes/COGITO_3_Lobby.tscn` same for Legacy demo

OvercastInteractive:
Just a quick look from me as well.

Prior to resolving the issues that @niefia pointed out, I did come across an error where input action "pause" didn't exist, which must have been a custom legacy input action used prior to using the built-in "ui_cancel".

Cogito_Input_Action_Pause_Not_Found

I'll continue testing it out this weekend.

OvercastInteractive:
Scene references that I needed to reconnect:
MainMenu.tscn -> MainMenu_SaveSlotManager -> `start_game_scene = res://addons/cogito/DemoScenes/COGITO_1_LegacyDemo.tscn`
MainMenu.tscn -> LegacyDemoSceneSwitcher -> `start_game_scene = res://addons/cogito/DemoScenes/COGITO_1_LegacyDemo.tscn`
PauseMenu.tscn -> MenuSceneSwitcher -> `main_menu_scene = res://addons/cogito/DemoScenes/COGITO_0_MainMenu.tscn`

Phazorknight:
@niefia / @OvercastInteractive 
Thanks for catching these! These should now be fixed.

niefia:
The only other error im seeing is this:
image
which should be fixed by replacing `menu_template_manager.gd` with `menu_template_manager.tscn` in autoload

niefia:
Can't seem to find any issues with it at this stage, so technically looks good to merge now.

However should probably also make sure all of the moved files match their latest versions on main to avoid losing any progress, 
and put a temporary pause on any other PR merges at this stage so that nothing gets lost in the move.

Phazorknight:
> However should probably also make sure all of the moved files match their latest versions on main to avoid losing any progress,

Agreed, will do one more commit now that grabs the latest from main.

niefia (approved):
Runs perfectly as expected, so looks good to merge now.
Haven't been able to manually verify every change obviously, but since it runs fine I'd say that's good enough. 
If there are any problems introduced with this restructuring I haven't been able to find them.

PR #313, opened 2024-10-15
Added Currency check to Button

niefia:
Button currency check added to function as a vending machine

PR #314, opened 2024-10-15
Currency Check component

niefia:
Rework of #313 to comply with new structure.

Instance of Currency check component has been added as a child of the button


Phazorknight (approved):
Thank you! Looks good to me, just thumb up this comment if you're done and I'll merge it in.

PR #316, opened 2024-10-15
Legacy scene fixes

niefia:
Realised legacy scene hadn't been checked with restructure, so tested it & found issues with scene paths and locked door being outdated.
 Fixed these to allow legacy scene to function still.


Phazorknight (approved):
Nice, thanks for this!

PR #322, opened 2024-10-25
Initial attempt at a basic vending machine, located in the breakroom

OvercastInteractive:
- Added a vending machine scene, which utilizes the existing `cogito_button` and `CurrencyCheck`
- Added `cogito_vendor.gd` which handles spawning, manages its own inventory and provides interaction hints
- I moved some code in cogito_button.gd from `interact()` to `press()` which is both more appropriate and prevents a bug in which I would be charged for a transaction without the `pressed()` signal ever being sent because they were handled independently. In short, the transaction only goes through once `cogito_button.gd press()` is actually called
- I also use a saved variable called `is_dispensing` which allows the player to save the game while dispensing of the object is running a timed delay. This means when loading the game, any items that hadn't yet been dispensed (saving and quitting prior to dispensing the item) will still be dispensed.


BrodyB (changes requested):
This is looking solid!

However, I do notice a soft error on picking up a health potion with this update. This is because the 2D health potion sprite auto-converts to a 3D texture at runtime because it's being used on a 3D Sprite node in the vending machine scene. 

I have a fix for that on my own version of `vending-machine`, which I can push if I have change permissions (I removed the sprite and made a jokey soda brand with text), or you can do a similar change.

It's a very minor thing, but probably best not to introduce any errors at all.

OvercastInteractive:
Hey thanks @BrodyB! I couldn't figure out how to grant you change permissions 😖 so I went ahead and made the changes. Let me know if you spot anything wrong with it.

BrodyB (approved):
Looks good to me!

Phazorknight (approved):
Fantastic work, thanks so much for this!

PR #332, opened 2024-11-21
Removing Cogito Editor Tab, instead using globals + custom resource CogitoSettings

Phazorknight:
This PR will get rid of the Cogito Editor Tab and it's options.

Instead, Cogito will use a new global/autoload script called `cogito_globals`, which loads a new custom resource CogitoSettings.
As a lot of script referenced `CogitoMain` and now instead reference `CogitoGlobals`, this PR touches a lot of files.

I commented out the  `EditorInterface` line that caused the issues on exported builds. I kept the function that creates a default project input map in, in case we find a way to implement it in a better way in the future.


OvercastInteractive (approved):
I had no issues with the vanilla settings windows .exe build when testing various scenes, saving/loading, multiple input devices, inventory and cogito object interactions.

Phazorknight:
@OvercastInteractive Thank you!

PR #335, opened 2024-11-25
Expanded upon throw and drop behavior

OvercastInteractive:
I made a few changes to `PlayerInteractionComponent` that allow for finer tuning of throw and drop power, based on the carried object's mass and limited to a max power.
- Using the object's mass in the throw and drop power provides more consistent and less chaotic behavior as very lightweight objects won't be launched at high speeds through the level and very heavy objects will only experience force up to max power.
- Also included the option to drain stamina if throwing objects that exceed a specified throw power threshold.


Phazorknight (approved):
Very nice, appreciate this mechanic getting some tweaks and refinement.
Thank you!

PR #340, opened 2024-11-28
Adds an AutoPickUpZone to the player

OvercastInteractive:
- Allows you to automatically pick up select items for whatever reason (frustrating, tedious, required, or to speed up the pace of the game) and may provide a smoother experience than only using the default raycasting + input interaction for every pick up.
- Adds a pick up delay to cogito projectiles, so they aren't automatically picked up the moment they're fired.
- Added laser ammo and foam bullets to the auto pick up list for testing.
- It's slightly counter to immersive sim in theory, however I feel like there are enough use cases that it could at least be optional.


Phazorknight:
Oooh, this is cool.
I would've probably tried to add this to the `PickupComponent`, but I think your implementation actually has some advantages (better control over pickup range + ability to deactivate it for certain moments like cutscenes etc).

I had a weird issue and also a crash when testing this:
1. Add some `projectile_pistol` to the Laboratory Demo scene and start the game in it.
2. When walking up to it, for some reason, the `projectile_pistol` doesn't get picked up by the auto pickup.
3. When I then walk over to the `pickup_laser_ammo` on the counter, the game crashes with an error:
```
player_hud_manager.gd:
Line 180: if node.is_disabled:
```
_Invalid access to property or key `is_disabled` on a base object of type `previously freed`._

Seems like there might be a bit of a race condition problem with freeing up these nodes at times.

I've also tried to remove and re-add the projectiles from the auto pickup, but still wasn't able to get them to get picked up by it.
Was able to add the batteries to it and they worked fine.

OvercastInteractive:
Thanks for testing this out a a bit more! I'll look into the issues you've had and see if I can safeguard the implementation against these kinds of errors. It may be a day or two before I can do a once over on this tho.

Phazorknight (approved):
Hey, gave this another test. Seems to be much improved! I wasn't able to replicate the errors I previously flagged, so happy to merge this in!

PR #345, opened 2024-12-05
Fix head's position offset in stair stepping by using only y component of the result

mrezai:
It seems there is a bug or limitation in Physics Engine when collision shapes change in runtime that cause result of body_test_motion to be unreliable, for example in current implementation of stair stepping we used body_test_motion to check motion only in y axis but in the result, by calling get_reminder or get_travel functions we get values that x and z are non-zero too.
This video shows what happen when bug isn't fixed and after some crouch/auto-uncrouch in window, camera goes through walls because head offset changed in x and z wrongly:

https://github.com/user-attachments/assets/9eea2f70-e355-4940-930e-b2aaedfb570a


Phazorknight (approved):
Makes total sense.
Thanks for this!

PR #365, opened 2025-02-06
cogito_button: Added class_name CogitoButton to allow scripts to inherit it

BrodyB:
I have a project that creates a button that inherits from cogito_button, but to do this, cogito_button needs to have a set `class_name` property.

I also changed the `pressed()` signal to pass a reference to itself. This is handy in cases where you have an object (say: a vending machine) and you get references to all the buttons in a for loop and connect them to a central `_on_button_pressed` function. Then when a button is pressed, you can determine which button was pressed.

### DIFF
--- a/addons/cogito/CogitoObjects/cogito_button.gd
+++ b/addons/cogito/CogitoObjects/cogito_button.gd
@@ -1,5 +1,6 @@
 @icon("res://addons/cogito/Assets/Graphics/Editor/Icon_CogitoButton.svg")
 extends Node3D
+class_name CogitoButton
 signal object_state_updated(interaction_text: String) #used to display correct interaction prompts
 signal pressed()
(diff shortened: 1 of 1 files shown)


Phazorknight:
Makes total sense, though changing the signal to pass the button breaks functions currently connected to it.
(eg the spawnzone in the Laboratory)

We could either update the affected scripts to accept the passed CogitoButton, or make it a separate signal... 🤔 
I personally lean towards a separate signal (as it avoids breaking stuff when people pull the update), but not sure what I'd name it.

BrodyB:
I've changed `pressed()` back to a signal without params, and added in `pressed_ref(button: CogitoButton)` as a version that passes the `CogitoButton` reference upon pressing.

Phazorknight (approved):
Thanks for updating this!

Looks all good on my end!

PR #367, opened 2025-02-11
SDPC states charted

Vroodot:
While not all completely functional, most Movement States for the player's current controller have been mapped out.

* Abstracted Sprint and Move states
* Gave each state a Class Name
* Made @export States Type Safe
* other minor (hopefully inconsequential) tweaks to SDPC
* Updated the Test Area with full State Chart template

NOTE: Not all of the states are functional/do anything. This pass was mostly to just get the states mapped out


Phazorknight:
Thanks for this! Makes sense to get the foundation laid out, looking forward to future updates.

Vroodot:
Nothing worth noting on most recent Commit.

I just wanted to ensure GitHub Desktop was working the way I assumed.

Phazorknight:
Thanks for this work so far. Looks intuitiviely structured.

When pulling and running this, had an error where no _PauseMovementState_ was assigned. I quickly assigned the SDPCStateInterfacing (might wanna rename this to _PausedMovement_ or something similar that's a bit more intuititive.

I think some functionality could probably be combined under one state instead of splitting it up too much but I asssume we'll discover that as we work our way through it. For example I'd say 'bunnyhop' logic can be part of the 'jump' state.

Vroodot:
No worries!
Good catch on the Interfacing/PauseMenu thing!
That was something I'd planned on updating once I discovered the distinction between `paused` and `ui_open`but guess slipped through cracks. Will be fixed on next push.

Will experiment with size of States and get back to you.

Here's my current reasoning on the granular implementation using Bunnyhop as an example:

Since this State can only be accessed through sliding, it can be in a clean chain
`Idle -> Move -> Sprint -> Slide -> Bunnyhop -> [Chain Bunnyhop Tech]`

While Jump itself is already kind of expecting a lot of Inputs:
i.e: Stamina Level, On Ladder, Climbing, Sprinting, Moving, Standing, Crouching

Having certain niche chains like this would keep specific movement tech from muddying up the rest of Jump Script.


However, I did a quick mockup of your suggestion.

psuedo-code
```
on Jump:
 if was_sliding:
    # Bunny Hop Stuff
 else: 
    # Regular Jump Stuff
    
on Landing:
  if was_bunnyhopping:
    # Bunny Hop Chaining Stuff
  else:
   # Regular Jump Landing Stuff
```

And I'm pretty sure I can work around either method.
Just let me know your preferred way forward 👍

And no worries! It may be a couple of days until my next update, but I'll be in touch.

Issue #362, opened 2025-01-24
player goes lightspeed when bunny hop is deactivated


limflag:
hi!
was making some tests on the framework and discovered by accident a bug where the player if sprint then jump, with the bunnyHop disabled, the speed after landing just goes insanely high
checked some codes from the player controller and found that may be that

![Image](https://github.com/user-attachments/assets/309c534e-5b92-475f-b2ff-4b95b7683324)

Cogito 1.0.2 and Godot 4.3 running on a Ubuntu 24.04

**Description:**
Looks like the SPRINTING_SPEED variable increment itself to a point where the player become faster than it should be

**Reproduction steps:**
Disable the player bunnyHop
Tries to sprint and jump

**Expected behavior:**
Player will be in a super high speed mode, like my SPRINTING_SPEED var got to 24, was set to 6 before

**Screenshots:**
uploaded before in first paragraph

https://github.com/user-attachments/assets/dac213e5-0578-452c-b6d1-aa9d95f3550f

limflag:
kinda fixed it when i removed the increment on the elif clause
```
   elif:
       SPRINTING_SPEED = SPRINTING_SPEED
```
probably just removing the clause should work too

Vroodot:
Added this as a small PR for the main build since its only a bug-fix 👍

I'm helping work on an updated Player Controller, so if you note any other issues please feel free to let me know

Vroodot:
@Phazorknight 

Since you accepted the PR, feel free to close this issue whenever

PR #368, opened 2025-02-12
Bug fix for current build: Update cogito_player.gd

Vroodot:
Commented out the cause of the bug from issue #362
Small two-line adjustment

Since we aren't adjusting the `SPRINTING_SPEED` at any point, there's no need to add it back to itself.

If adding a Jump Acceleration is the desired behavior, I detailed how we could achieve that in the comments using a similar method as `bunny_hop_speed`


Phazorknight (approved):
Seems all good from my end!
Thanks!

PR #398, opened 2025-04-22
Added loot generation functionality

vianuer:
New component: Loot Component. Place it under any node that has a valid health component and configure the options.

This component handles the loot generation upon entity death within a CogitoContainer. A new PackedScene called loot_bag.tscn can be used an example. 

Simply add it under any node that has a health component and configure it in the editor. Loot table examples can be located to create your own loot tables for specific loot generation.

v1 currently supports multiple items in loot table, rolls random items based on weight, drops it in the loot_bag after generation and handles the loot bag despawn in three different ways. 

There are some modifications to core that is required for advanced functionality for unique items. Also the laboratory scene has a corner dedicated to testing loot components functionality with three barrels. Component also supports quest item drops with quest id comparison during drop generation. 

A more detailed manual on functionality should be available later. 

Known issues: Timed despawn functionality listens to toggle_inventory signal which pauses and continues a destruction timer. If the inventory interface is closed with escape key, signal is not triggered and timer does not continue.


Phazorknight:
Cool, I'm quite busy this week, but hope to be able to review this on the weekend.

Phazorknight:
This seems to work pretty well!

Noticed an issue where if I had the inventory open while the self-refilling container refilled, it's inventory would pop up in the inventory screen. Probably need to disconnect some signal to avoid this.

vianuer:
Yeah I've noticed that too. Hoping to come with a fix for that soon. 

Also maybe unrelated but noticed a save reload bug with the new NPC class. If you save in laboratory map with npc alive and try to reload a null reference comes up. Killing the npc and saving doesn't create the issue. I can create an issue in detail if requested.

Phazorknight:
> Also maybe unrelated but noticed a save reload bug with the new NPC class. If you save in laboratory map with npc alive and try to reload a null reference comes up. Killing the npc and saving doesn't create the issue. I can create an issue in detail if requested.

If you could create an Issue report, that'd be great. I haven't done a lot of testing with the CogitoNPC yet.

vianuer:
I'll create an issue when I've done some more testing on it.

I've been doing some testing with save and reload to confirm loot container functionality. What would you recommend for creating persistent timers? Since lootable containers which respawn can be spread out through other scenes, their respawn timer currently starts over when they are loaded into the scene again. A way to save and reload these would be great.

Phazorknight:
> I've been doing some testing with save and reload to confirm loot container functionality. What would you recommend for creating persistent timers? Since lootable containers which respawn can be spread out through other scenes, their respawn timer currently starts over when they are loaded into the scene again. A way to save and reload these would be great.

Any CogitoObject gets saved with it's own properties saved as a dictionary. If you add the time remaining as a property that gets saved to their class this should be easy to save.
So in the `save()` function add something like:
`"spawn_time_remaining" : spawn_time_remaining,`

then make sure in their `set_state()` function that property is loaded, something like:
```
if spawn_time_remaining > 0:
  timer.wait_time = spawn_time_remaining
```

This is mostly pseudocode, I haven't compared this to your actual scripts.

vianuer:
Phew that took a while to iron out. Saving and loading should work as intended and with the added bonus that you can make sure that timer can or can not count down while unloaded. 

Hopefully with time this can be implemented. I will see if I can actually make the spawned items stack during generation. Beyond that if you can think of anything else that can be useful, I'd be glad to hear it.

Phazorknight:
Thanks! Should be able to review this before Monday.

Phazorknight:
@vianuer  This seems great, fantastic work!

Only have two notes:

> Noticed an issue where if I had the inventory open while the self-refilling container refilled, it's inventory would pop up in the inventory screen. Probably need to disconnect some signal to avoid this.

This issue still seems to be there, occurs with the fridge that respawns its contents.

**Despawning Logic**: Love these options. For the _empty_ option, would it be possible to implement a flag for the container to only despawn on closing the inventory screen?
Mainly as if I play through this, I press "interact" to open the container, but once it's empty and it despawns, I can no longer press "interact" to close it (have to use "Tab" instead). So I'd love an option to keep the container around so the same button closes it.
Hope that makes sense!

vianuer:
Hey I am glad you liked it! I enjoyed working on this. The issue happens when you close out of the container with Tab or Esc keys, normally closing should be ok. I'd have to check signals more and maybe do a few more modifications to core to fix that issue. 

Oh yeah I understand what you are referring to, I remember noticing how jarring it was when container suddenly disappears but I argued in my head that most people would probably use the timed despawn option but I will make sure that functionality is included.

vianuer:
It is ready for another review, I've implemented the functionality you've requested and without doing any pesky core edits.

vianuer:
Oh I forgot to delete the loot_barrel, it is not used anymore. Now the debug prints all pipe into debug_log and debugging functionality is enabled in cogitosettings. Added some more lines to manual. 

You may review it again whenever you like :)

Phazorknight:
Nice, getting there!

Ran into two issues in this commit:

`@export var debug_prints: bool = CogitoGlobals.cogito_settings.loot_drop_container_debug`

these (and the similar lines) gave me errors. Since you're referencing the settings resource, changing them to @onready would probably make more sense.

Secondly, I ran into an error when switching back into a sceen with spawned loot containers.
Repro steps:
- Run currents scene Laboratory
- Shoot the targets to get the loot containers to spawn
- Leave level through the door to enter the lobby.
- Turn around and return to the Laboratory
- Game errors here on the CogitoSceneManager, line 316 - this is where saved nodes get instantiated from their packed scene path.

image

vianuer:
Oh yeah keeping debug_prints as export wouldn't make much sense anyway, thanks for reminding me. Should be addressed with 02a552d. 

Error is because I've apparently forgot to merge the changes you've made in #402 and it was early NPC save and load behavior that was causing the problem. Should be fixed now. I've tried it again after syncing the changes with a8f3916 and no crashes happened.

Phazorknight (approved):
Amazing work.
Would love to see grid inventory implemented in the future, but I'd consider this good enough for now.

Also had a question:
Would part of this be usable to do random drops of pickup-items (not in a container, but directly spawning them?)

So for example, if shooting the targets, could they randomly drop a potion, gold, etc based on a loot table?

vianuer:
I am glad I could contribute. I was actually trying to implement something like that, while researching I came across the #403. I was trying to achieve a loot explosion like when you slice enemies in doom with chainsaw and they pop 10 maybe 20 ammo and health drops but picking up the darts I was spawning was causing freezes. I will implement that loot drop version too but probably without the loot explosion part. Instead of spawning a container and fill it up, that setting would spawn a pickup version of the item or maybe a custom class like you've described in #403.

As for grid inventory, loot container can be expanded to any size during creation, it would need to account for the total inventory size. I haven't checked how the grid inventory is set up in core. I'll look around and see if I can tackle that but can't really make any promises for grid inventory since it looks like magic to me at this stage haha.

PR #406, opened 2025-05-12
Implemented new spawning logic for just item spawning for the loot component

vianuer:
- Added spawning logic to either spawn a container or the item defined in drop_scene from InventoryItemPD.

- Added a small impulse to spawned items so they don't bunch up.

- Fixed debug variables pointing to wrong option, now they correctly toggle with their intended variables.

This way of spawning items is much more interactive and entertaining. I had to stop myself from filling the entire room with potions. However these items are not picked up by autopickup component. Player has to manually clean all this up.

image

### DIFF
--- a/addons/cogito/CogitoObjects/cogito_loot_drop_container.gd
+++ b/addons/cogito/CogitoObjects/cogito_loot_drop_container.gd
@@ -2,7 +2,7 @@ class_name LootDropContainer extends CogitoContainer
 
 
 ## Enables the debug prints. There are quite a few so output may be crowded.
-@onready var debug_prints: bool = CogitoGlobals.cogito_settings.loot_generator_debug
+@onready var debug_prints: bool = CogitoGlobals.cogito_settings.loot_drop_container_debug
 
 ## Determine the despawning logic for this container.
 @export var despawning_logic := DespawningLogic.NONE:
--- a/addons/cogito/CogitoObjects/cogito_lootable_container.gd
+++ b/addons/cogito/CogitoObjects/cogito_lootable_container.gd
@@ -1,7 +1,7 @@
 class_name LootableContainer extends CogitoContainer
 
 ## Enables the debug prints. There are quite a few so output may be crowded.
-@onready var debug_prints: bool = CogitoGlobals.cogito_settings.loot_generator_debug
+@onready var debug_prints: bool = CogitoGlobals.cogito_settings.lootable_container_debug
 
 ## Determine the despawning logic for this container.
 @export_category("Execution Configuration")
--- a/addons/cogito/CogitoObjects/cogito_object.gd
+++ b/addons/cogito/CogitoObjects/cogito_object.gd
@@ -12,7 +12,7 @@ signal object_exits_tree()
 var interaction_nodes : Array[Node]
 var cogito_properties : CogitoProperties = null
 var properties : int
-
+var spawned_loot_item: bool = false
 
 func _ready():
 	self.add_to_group("interactable")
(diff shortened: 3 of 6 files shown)


Phazorknight:
Now thats what I'm talking about, haha.
This is a lot of fun.

Edited a loot table and noticed two things:
- Is it possible to set up a table that has a chance of dropping nothing at all? Wanted to see if I could just add coins with a chance of 20%, but still every destroyed spawned coins.

- I don't expect this for this pull request, but I noticed that setting up new entries in the loot table is a bit tedious with the dictionary. Would it be possible to create a custom resource for this?

vianuer:
Loot tables are indeed tedious to handle by hand however this is the resource version as far as I am aware. I am not sure how I could streamline the loot tables a bit more. If you have any idea about how to set this up, I am all ears. 

For a loot table to drop nothing I would have to do a bit of editing in the code, I recall specifically programming the loot component and loot generator against not doing empty operation in loot tables because null is bad. However a more elegant way of handling it could be implemented, say by making the drop itself a chance based event. Maybe it drops based on a percentage or always? Either way I will work on this. 

Seems I've forgotten to include the quantity bit when spawning the objects. I've added it in the next iteration but ran into an issue of displaying quantities on the items themselves. I am editing the display_name property of slot_data of the spawned cogito object, this is proven to be unreliable, sometimes it does nothing, sometimes it edits all the copies of the item even in inventory. A more elegant way of displaying would be by done in something like UI_ObjectNameComponent I think. This would also come in handy if we were to be able to drop the item stack in its entirety aside from dropping a single copy. My finger is still tired from pressing G 100 times haha.

Phazorknight:
> Loot tables are indeed tedious to handle by hand however this is the resource version as far as I am aware. I am not sure how I could streamline the loot tables a bit more. If you have any idea about how to set this up, I am all ears.

I might give this a go on the main branch so we don't cross wires, but will prob just be a single script that will replace the dict array.

> I am editing the display_name property of slot_data of the spawned cogito object, this is proven to be unreliable, sometimes it does nothing, sometimes it edits all the copies of the item even in inventory.

This might be the classic "Local to Scene" pitfall, if not set to on, reources are treated as one instead of properly instanced. So making sure to flag this to _true_ when spawning might help.

Phazorknight:
> I might give this a go on the main branch so we don't cross wires, but will prob just be a single script that will replace the dict array.

@vianuer , I gave this a go in https://github.com/Phazorknight/Cogito/pull/407, please take a look. Working on #407 and this #406 in parallel might cause some conflicts, but I don't think it'll be too hard to fix. Happy to give #406 priority.

vianuer:
I think we can port the changes in #407 to #406 easily as there aren't major changes to code and those changes would be much more beneficial in the long run. There a few things that keep cropping up like how loot popping isn't handling quest or unique items at this time.

Oddly enough I've set it up as the battery only gets a differing quantity per drop and the item itself is set to local in the resource itself. I will make sure that spawning script does set the items spawned to local regardless.

vianuer:
There a few things to know regarding loot popping. Quest and unique checks rely on groups items placed during spawn. These groups are not saved during save and load functions so in 900d2bd CogitoObject class had to be modified to save the group entry regarding the spawned_loot_items group. 

If you've created a lot of items with the target destructables you should have ran into this issue. You can literally fly off if you sprint over the items and other loot objects can fly everywhere too. While it is entertaining, loot can get stuck between other objects, drop off the floor to Narnia etc. so not exactly stellar gameplay experience. I think some modification to item collision layers is necessary so loot doesn't interact with player's collision shape. Removing the object from layer 1 and moving them to its own layer in say layer 3 makes them less prone to flying around but I am not implementing this yet.

Phazorknight:
Great catch on the save/load issues - and a solid solution for it. 👍🏼 

> While it is entertaining, loot can get stuck between other objects, drop off the floor to Narnia etc. so not exactly stellar gameplay experience.

I see this as a separate physics issue. Physics layers could be set on the pickup object scene directly.

Phazorknight (approved):
Looks good! No major issues encountered when testing.

PR #407, opened 2025-05-13
Change LootTables to work with LootDrop resource (instead of dictionaries)

Phazorknight:
*Re: Loot Table entries as resource*

So I gave this ago and I feel I'm pretty close to getting it to work, just having some trouble adapting the existing code as I'm quite unfamiliar with it.

- Created a new resource called `LootDrop.gd` (very open to change the name, but that seemed to make sense to me as you fill a table with drops. Looks very straightforward, just adapted your example dict entry.

- Changed _BaseLootTable.gd_ to accept `Array[LootDrop]` instead of Dictionary.
Pretty much just looks like this:
```
class_name LootTable extends Resource

@export var drops: Array[LootDrop] = []

func _init() -> void:
	pass
```

- Then I changed pretty much any reference to LootTable dict arrays to be Array[LootDrop].
- Then I changed any of the loops etc. to access the lootdrop properties directly instead of getting dict entrys. This is pretty easy overall, but I had to change the `cogito_loot_generator.gd` logic handling quest items a bit (as every LootDrop has that property), so instead of checking if the property exits, it checks the value.
```
162		# Handle Quest Items			
163		elif _winner.quest_id > -1: # If winner is a quest item
```

It seems there's still some snags or I missed updating the existing scenes and resources to use this, but wanted to make this public so it can be improved/reviewed.


vianuer:
Hey I've just tested this out, implementation is great, actually user friendly, not only you can save the drops for in a directory should you wish, you can also just fill them out on individual loot components which is what I intended initially. Improving this additional properties shouldn't be a headache as it was with the earlier loot table concept, should be safer too. 

I've tested the quest item spawning logic as you've mentioned, works as intended. Currently I can't finy any problems with it. LootDrop is alright but I think it may be more informative to call these as a LootDropEntry since a loot table is nothing but an array of loot drop entries in essence.

Phazorknight:
> LootDrop is alright but I think it may be more informative to call these as a LootDropEntry since a loot table is nothing but an array of loot drop entries in essence.

Makes sense to me! Just updated this.

I also did a bit more code cleanup and updated existing tables to use the LootDropEntries as well.
If you think it's good, I'd merge this into main, and then we could probably update #406 with this.

Phazorknight:
> So a null check for the _item_count fixes that error and just drops an empty container

@vianuer  added! Thx for the catch.

Phazorknight:
@vianuer I assume this is good from your end.

As I can't review/approve my own PR, I've flagged this to the other contributors.
@vianuer if you're interested in becoming a contributor, let me know and I'll add you.

vianuer:
Of course, I'd be happy to help. I've tested the loot tables, with both unique and quest items so far everything works as expected, if something were to come up I'd PR a fix for it down the line for it.

vianuer (approved):
Seems to work as expected. I will approve the changes.

PR #415, opened 2025-05-30
Allows automatic quickslotting based on ItemPD type

OvercastInteractive:
I find this quite convenient over the standard procedure for quickslotting, although it may not make sense for some projects.

- Added a signal to cogito_inventory when picking up an inventory item that requires an empty inventory slot, which connects to CogitoQuickSlots.
- Added booleans for each ItemPD subclass to determine what types of items can be quickslotted. This could be changed to an array of specified item data for precision.
- When the 'new' item is picked up CogitoQuickSlots will attempt to automatically quickslot it.
- Known Issue: when transferring items from player inventory to external inventory the quickslot icon isn't cleared, which occurs with and without this commit's changes.

Notice the quickslot binding with the item automatically and then being used without ever interacting with the inventory:
Auto_quickslotting


mrezai:
Based on all above discussions, I think for most flexibility we can add these to InventoryItemPD and change current implementation accordingly. Preferred concept removed from this design:
```
@export_category("Auto Quickslot Settings")
@export var can_auto_slot: bool = false
@export var slot_index: int = -1
```
- can_auto_slot = false, inventory and quick slots work as before.
- can_auto_slot = true and slot_index = -1, item quickslotted in first empty slot if there is one.
- can_auto_slot = true and slot_index = 1, item quickslotted only if slot 1 is empty, else no quickslotting.

I think this feature mostly useful for items like wieldables, flashlight and maybe some potions, other items can have default settings.

OvercastInteractive:
Thanks for laying out the preferred approach @mrezai

> * can_auto_slot = true and slot_index = 1, item quickslotted only if slot 1 is empty, else no quickslotting.

I renamed 'slot_index' to 'slot_number' and have a value of 1 refer to slot 1.
If devs would rather use slot_index then this should be changed sooner rather than later.

Also, I noticed that some InventoryItemPD resources weren't referenced from their asset folders and were instead created as new resources in the PickUpComponent in their 'pickup' scene, which meant that they didn't receive changes to the resource, if that makes sense. This could lead to headaches when changes you make to an item's InventoryItemPD resource have no effect in the item's pickup component. This happened to me with the Laser Rifle and some other items.

mrezai:
Looks good to me.

> Also, I noticed that some InventoryItemPD resources weren't referenced from their asset folders and were instead created as new resources in the PickUpComponent in their 'pickup' scene ...

I found another issue, I added Cogito_LaserRifle.tres to "Auto Pick Up Items" array in player's AutoPickUpZone to test auto pick up and auto quickslot together but it doesn't work. In this line ```pick_up_component.slot_data.inventory_item``` has different resouce id for laser rifle each time but for flashlight it has same value.

https://github.com/Phazorknight/Cogito/blob/ac0298e05286b2b6d9ba7c312833540037302e5a/addons/cogito/Components/AutoPickUpZone.gd#L85

OvercastInteractive:
When you make a resource Local To Scene comparing inventory_item data breaks. When you mark wieldable pickups as Local To Scene it allows having multiple wieldables of the same type, such as two different Toy Pistols, function properly in inventory and in quickslots (thanks to @Phazorknight for helping me with this!). Because of this I've been comparing items by name (this.inventory_item.name == that.inventory_item.name) with the assumption that items of the same name derive from the same item data.

A quick fix is to append all inventory_item.name to an Array[String] from the auto_pick_up_items array and use string comparison as matching data won't work with Local To Scene checked. There's likely a better solution though!

Phazorknight:
> A quick fix is to append all inventory_item.name to an Array[String] from the auto_pick_up_items array and use string comparison as matching data won't work with Local To Scene checked. There's likely a better solution though!

I'm honestly not sure there is.
Checking directly for resources comes with caveats, as @OvercastInteractive points out - another thing is that when assigning resources in editor, you can run into Godot blocking you from creating "cyclical references".
I ran into this when creating ammo items (hence wieldables referencing them by name).

Back to the PR, this seems to work great, and seems easy to use!
Some code cleanup and some added inspector hints, then I'd be happy to merge this in.

OvercastInteractive:
I'm not sure what else to add to this PR that others deem important. Let me know what you'd expect as far as code cleanup and inspector hints, or feel free to merge this first and make the changes if you it's not too much trouble!

Phazorknight:
> I'm not sure what else to add to this PR that others deem important. Let me know what you'd expect as far as code cleanup and inspector hints, or feel free to merge this first and make the changes if you it's not too much trouble!

Sure, I can do that, no worries! Thanks again!

PR #420, opened 2025-06-05
Fix the possibility of adding non-CogitoObjects to pickup_item_pool

mrezai:
Fixes this error:
Screenshot from 2025-06-05 10-48-14


Phazorknight (approved):
Eeyyyyy! Thanks so much for fixing that!

PR #422, opened 2025-06-07
Allows PickupComponent and BackpackComponent interaction while carrying an object

OvercastInteractive:
- Checks the carried object for a PickupComponent or BackpackComponent, and if present, shows the input prompt and allows interaction with these components.
- A quality-of-life enhancement that retains the player's ability to pick up the carried object.
- Rearranged the inventory_bag scene children such that the BackpackComponent is childed above the CarryableComponent, because the input prompts are shown in the order they're childed (while not carrying an object).
- This feature makes the most sense when it comes to items that can be both placed in inventory and used in Snap Slots where you're expected to carry objects into snap shapes but may decide to put them back into inventory instead.
- Without this feature you must drop the object (somewhere safe) and then look down at it to pick it up, or worse, drop it and then quickly pick it up while it's falling. This is a bit inconvenient and not very immersive.
- This implementation could certainly be reworked, I just wanted to change as little as possible.

https://github.com/user-attachments/assets/bbec6163-0c8d-44c7-8a3a-0560e4a0af48

### DIFF
--- a/addons/cogito/Components/PlayerInteractionComponent.gd
+++ b/addons/cogito/Components/PlayerInteractionComponent.gd
@@ -108,13 +108,23 @@ func _input(event: InputEvent) -> void:
 func _handle_interaction(action: String) -> void:
 	# if carrying an object, drop it.
 	if is_carrying:
-		if is_instance_valid(carried_object) and carried_object.input_map_action == action:
-			_drop_carried_object()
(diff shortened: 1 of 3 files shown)


Phazorknight:
Oh nice! Will take a look at this.

Phazorknight (approved):
Thanks for this!

PR #423, opened 2025-06-10
Improve dropping inventory's items.

mrezai:
- These added to player: radius variable and _calculate_player_radius function.
- In shape_cast, sphere replaced by box shape.
- AABB added to CogitoObject for bounding box checking. If Pickup has MeshInstances, get_aabb function will be used to calculate AABB. It's possible to set a custom AABB using custom_aabb property for complex meshes or CSGs.
- If there isn't enough space to drop item, error sound will be played.
- It seems all pickups are using cogito_object.gd except pickup_diamond_key_complete.tscn that uses cogito_pickup.gd so AABB doesn't work for this item.

@Phazorknight please review this and if this is an acceptable approach, do final necessary modifications, thanks.


Phazorknight:
Wow, this is pretty cool!

Two things from my end:

1. When dropping an item is blocked due to lack of space, I expect there to be a hint message about it (not just the error beep). Want to make it clear why we block the dropping of items.

2. Setting a custom AABB in the inspector is very unintuitive, so I took the liberty and added in a debug draw option. Can probably be polished up a bit but should work and help to set those AABB shapes (and checking if the auto-detected ones look correct).

@mrezai let me know what you think / feel free to add and tweak.

mrezai:
I will test these changes soon.
> I've changed this into a bool now. As long as the bool flag is turned off, no debug shapes should stay visible.

I think we can use this instead of bool flag:
https://docs.godotengine.org/en/4.4/classes/class_engine.html#class-engine-method-is-editor-hint

PR #424, opened 2025-06-10
Fix auto pickup for items that use Local To Scene.

mrezai:
Fix is based on this https://github.com/Phazorknight/Cogito/pull/415#issuecomment-2932499189


Phazorknight:
Hey, thanks for this! 
Overall this works pretty intuitively.

One odd behavior I ran into though:
- I added the Health Potion and Coins to the AutoPickup.
- In the Laboratory scene, I shot the targets that drop coins and health potions.
- When running over them, I auto-picked up the Health Potions, but _not_ the coins for some reason.

I also noticed that the health potions that spawn can't be stacked with each other, but I think that's unrelated to this and should be a separate issue.

mrezai:
- Potion issue is fixed by same logic(check names).
- Coin interactable layer wasn't set.
- Hint UI's z-index set to -1 to show hint messages under opened inventory.

PR #428, opened 2025-06-13
Allows context-sensitive cycling through all quickslotted wieldables

OvercastInteractive:
- Pressing "interact2" (E on keyboard, top face button on gamepad) will cycle through quickslotted wieldables
- Don't worry, it's context-sensitive and only works when not detecting an interactable that uses the "interact2" input action
- Serves as a convenience feature found in most modern FPS-style games, allowing for quick cycling between wieldables using a single input button
- The E key is more convenient to press, can be pressed more intuitively, and for gamepads this gives the player the ability to select a quickslotted wieldable using either hand on the controller (d-pad with the left and top face button with the right)
- Very simple implementation in a single class and works with existing Cogito functionality to hopefully be bug-free

### DIFF
--- a/addons/cogito/InventoryPD/CogitoQuickSlots.gd
+++ b/addons/cogito/InventoryPD/CogitoQuickSlots.gd
@@ -5,7 +5,8 @@ class_name CogitoQuickslots
 
 ## If true, the player can assign the same item to multiple quickslots.
 @export var allow_multiple_quickslot_binds : bool = true
-
+## When multiple wieldables are quickslotted, cycling will unequip the last wieldable found before cycling back to the first wieldable found
+@export var allow_unequip_when_cycling_wieldables: bool = true
 @export var quickslot_containers : Array[CogitoQuickslotContainer]
 var player_interaction_component: PlayerInteractionComponent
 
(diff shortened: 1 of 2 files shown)


Phazorknight:
hey, haven't had a chance to check this out yet, just have been busy, hope to get to this over the weekend.

> To test, just pick up and quickslot between 1-4 wieldables, then press "interact2". You should cycle through them, and still be able to press their associated quickslot input action to change and equip them as you want.

Gotcha, I'd probably add new input map actions for this. I picture that the mouse wheel could be a great option (and to offer a cycle forward and cycle backward)

> At the moment, pressing "interact2" while wielding when only 1 quickslotted wieldable exists doesn't unequip it, which would probably be a minor, but useful, addition.

Yeah, that'd prob be nice - unequipping if the quickslot is empty.

> Also, an alternative to this quickslot approach would be to allow cycling through all wieldables in your inventory, which could potentially be even more useful but raises more question marks and certainly would involve trickier implementation.

I wouldn't expect this - plus I assume the cycle-order that players prefer might be different than how they have items arranged in the inventory. Would not recommend trying to implement this.

OvercastInteractive:
Just pushed changes that incorporate @PedlarGames mouse scroll wheel cycling functionality, @Phazorknight new input action events specific for cycling quickslotted wieldables to allow for rebinding and the commit also includes the option to unequip when cycling, which works for both mouse scroll and button events.

Let me know if you come across any issues!

PR #430, opened 2025-06-24
Added option for floor slope to affect run exhaustion speed and enhanced UI stamina attribute display behavior

OvercastInteractive:
- Added the option to use floor slope to influence run exhaustion speed. This works for steps too, by comparing the player's global y position over frames.
- This behavior lerps the base run exhaustion speed to either an uphill or downhill max exhaustion speed. The idea is that running uphill should drain stamina much faster, while running downhill should drain it a bit slower, either on ramps or steps.
- Enhanced the UI stamina attribute to be a little more flexible with how it displays, including whether it fades out while the attribute increases, fades immediately upon reaching max value, and blinking back on momentarily if faded out when reaching max so the player can know it has finished recovering.
- Fixed the stamina bar to be the same width as the background, like the other attribute component bars.


Phazorknight (approved):
Beautiful, no notes.
Very nice polish on this! Thank you!

PR #433, opened 2025-06-28
Added AutoConsumes, fixed loading from death screen and fixed a wieldable cycling bug

OvercastInteractive:
- AutoConsumes are assigned to an attribute and use attribute and inventory signals to check player inventory for items to automatically consume when an attribute crosses a threshold value.
- Typical use-cases are added, including an energy drink which is auto-consumed when stamina attribute reaches zero, and an emergency injector which is auto-consumed when health falls below one, essentially preventing death.
- The stamina extension potion was also included as a stamina auto-consume, so that it is consumed the moment it's placed in inventory. Some of this behavior could make sense as being variables in the ConsumableItemPD, such as when an item increases the max value, but I included it in AutoConsume as an example anyway.
- Also added a few lines of code to the death screen load behavior that allows using the "Load" button to work properly, as long as a save state exists (otherwise it still breaks)
- Also made sure the damage_overlay would be reset to transparent when loading from death.
- Added a check in CogitoQuickSlots to prevent wieldable cycling while carrying/dropping carriables.
- Changed Stamina Extension potion carryable drop input to "interact2" to be uniform with other carryable inputs.


OvercastInteractive:
> I agree that maybe we should implement the auto-consume feature into the ConsumableItem class, instead of a component to the player.

Perhaps as a subclass called AutoConsumableItemPD, or just add more variables to ConsumableItemPD. Instead of having a list of consumables stored in the component, just check for all AutoConsumableItemPD types in inventory. I don't yet know how I would implement this feature outside of using them as components to the player. They only reference the player's attributes so they could be childed under attributes but that wasn't necessary. I'll take another look all of this after I think about it some more. It was a bit of a challenge to get it working as mostly decoupled from the rest of the codebase.


> This would prevent potential issues like the Auto Injector item description being misleading if the player actually has no AutoConsume

Not sure what you mean. If the developer doesn't use the AutoConsume feature or the player doesn't have an AutoConsume in inventory? Feel free to change the description to whatever you like! There are many different ways to help identify an item's type/utility other than through language in the description, and all of that should be up to the developer to decide. 👍

> Can you confirm the source of your graphical and audio assets are CC0/MIT (and potentially add their credit to the credits page)?

Here are links and credits to the audio files, all are donationware that I also modified and mixed in Audacity:
(RibhavAgrawal) https://pixabay.com/sound-effects/spray-185000/
(VoiceBosch) https://pixabay.com/sound-effects/energy-drink-effect-230559/
(OxidVideos) https://pixabay.com/sound-effects/soda-can-open-183214/

> I think it's fine for this PR if that's not 100% solved, will look into adding some more safety check for loading saves further down the line in a separate commit/PR.

Yeah, it was a quick fix to speed up testing of the AutoConsume and I didn't expect it to work bug-free 😅

Phazorknight (approved):

PR #436, opened 2025-07-10
Added ExtendedPickupInteraction, HoldUi and update several scenes with fixes and improvements

OvercastInteractive:
@Phazorknight There are quite a few things here, with many smaller changes and a few larger ones. Take a look when you have the time and see if these hold any merit or need a re-think.

- ExtendedPickupInteraction: added to a few of the common Pickup items. It allows pick up types like Consumables, Ammo and Wieldables to be consumed/reloaded/equipped using the basic hold interaction functionality straight from the world rather than only through inventory.
- This also allows the player to quickly consume or reload the equipped wieldable even if they don't have space in inventory for the item. This sidesteps annoying situations where the player would need to drop an item (or more than one), then pick up the new item, then use it, then pick up the last item dropped again.
- UI_HoldComponent: holds all of the HoldInteraction UI logic as a dedicated class in the Player_HUD. This way, if you have many hold interaction objects they aren't all checking for input and processing the hold action. It works exactly as before, just not inside of the item's interaction logic (since you're only able to interact with a single object at a time, it makes more sense to do it this way).
- Add an interrupt timer to quickslot cycling based on interactable_unseen, which prevents the player from quickslot cycling momentarily after looking away from an interactable. If you try to carry an item but misaim and have quickslotted wieldables, this would trigger cycling between these when you were just trying to carry something.
- Added some textures to the ammo dispenser in the Laboratory.
- Added a vending machine and strength potion to the Legacy scene. The strength potion allows you to actually carry the carriable crate. The vending machine can dispense a couple more crates if you need one.
- Made DualHoldInteraction and ExtendedPickupInteraction extend from HoldInteraction, as that logically makes sense and simplifies both of their classes, along with moving their interface functionality to the UI_HoldComponent.
- There was a bug in the cogito_vendor.gd code that couldn't find the player or player interaction component on load so added an await.
- Updated the set_attribute function in cogito_attribute to set the max value first and then clamp the current value between zero and max.
- I also updated some animations to be a bit snappier and look a little nicer for the laser rifle and pistol.
- I tested this extensively in many scenes and scenarios and snuffed out every bug that I came across. It seems to work as intended and I think it's an improvement to the experience.


OvercastInteractive:
Another thing to note, I increased the player jump velocity to 4.5 because otherwise you can't jump up onto the carriable crate or get out of the walled-off room in the Legacy Demo scene at the previous jump velocity of 4.0.

Phazorknight (approved):
Thank you for this!
This all makes sense to me, really appreciate the polish on a lot of the details.

The ExtendedPickupInteraction is great, has that balance of being easy-to-use but also offering customziation.
Also really appreciate how you update the hold interaction (especially with the just added custom crosshairs.

Let me know if you still need to update/adjust things, otherwise I'll merge this in!

As a side note: I'll try to get the Localization PR ready soon (will update it to support this) so please give me a heads up if you're planning on more extensive PRs like this, then I might get you to hold off for a bit until the Localization feature is merged into main.

OvercastInteractive:
I have nothing more to change on this at the moment so feel free to merge it.

I'll hold off on PRs for a while and test any other changes I make on my Cogito-based project first!

Thanks for the heads up and looking forward to the Localization update!

PR #437, opened 2025-07-17
Add localization

Phazorknight:
### This PR adds full localization to COGITO.

It uses Godot's built-in POT generation, which has been updated to enable parsing of Cogito's custom resources, more specifically items and quests.

Additionally, there's a new custom resource called "Translation Key Dictionary" which can be used to store group of translation keys to be reused accross objects. For example, this way all doors can use the same translation keys, even if they are their own scenes etc, which eliminates the need to add all door scenes to the POT generation list and reduces doubling up of translation keys. At the same time, it enables the dev to customize the localization of specific objects. I've included a few translation key resources that are grouped by use.

**I've also created a quick video walk through that shows how this works in action:** https://youtu.be/MOUirXlXAfA
This should help anyone get started with localization quickly, even if they've never used it before.

We have so many strings and hints in COGITO that I'm sure I've missed one or two, but please take a look and see if this works as expected.
Due to updating a lot of the string composition, string references and hint messages, this PR changes a lot of scripts, but the changes within them should be relatively minor.

To test, open the in-game Options menu and switch the language (included are English and German)

### Known issues:
- Sitting: I was not able to update the interaction prompt of seats when a scene starts (it seems a bit weird in general, not always appearing). Once seated and getting up again, it updates as expected, but this is of course not there yet. Would appreciate any help in getting seats working.

### DIFF
--- a/addons/cogito/Components/PlayerInteractionComponent.gd
+++ b/addons/cogito/Components/PlayerInteractionComponent.gd
@@ -9,6 +9,7 @@ signal updated_wieldable_data(wielded_item: WieldableItemPD, ammo_in_inventory: 
 signal interactive_object_detected(interaction_nodes: Array[Node])
 signal nothing_detected()
 signal started_carrying(interaction_node: Node)
+signal update_crosshair(is_visible: bool)
 
 var look_vector: Vector3
 var device_id: int = -1  # Used for displaying correct input prompts depending on input device.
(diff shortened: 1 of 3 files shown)


vianuer:
Hey this is great! It will certainly come in handy. I haven't been to code side of things for a while trying to get a grip on art assets for my project but as my project is a text heavy one, I certainly appreciate the internationalization options. 

I have been testing the sitting problem you've mentioned, I think it has something to do with interaction text being accessed by both CogitoSittable and BasicInteraction components. From my testing BasicInteraction component overrides the interaction text as it is a child of the CogitoSittable. Also if you sit and try to stand up fast, you get to clip through the floor.

Phazorknight:
I was able to improve the "Sit" localization a bit.

Since I haven't heard anything, I'll probably merge this into main in the next 48 hours.

PR #447, opened 2025-08-17
Make carried object rotation more consistent

lorbog:
Changes carried object rotation so that it behaves consistently regardless of object's position or the rotation of the camera.  

Pressing forward will now always rotate the object away from the player, while pressing to either side will always rotate along the global y axis.


Phazorknight (approved):
Oh this is a no-brainer.
Thanks so much!!!

PR #450, opened 2025-08-27
Fix camera initialization in CarryableCompponent.

mrezai:
When player moves to a new scene, camera in CarryableComponent doesn't initialize correctly.


Phazorknight (approved):
Thanks for this! Makes sense to me!

Issue #454, opened 2025-09-03
Invalid access to property or key 'discard_after_use' on a base object of type 'Nil'.


arnoldblake:
**Cogito and Godot Engine Version:**
Godot 4.4.1, Latest

**Description:**
Opening the archive door with the keypad crashes the game with invalid access to a property on the key.

**Reproduction steps:**
Start a game, open the archive with the passcode

**Expected behavior:**
The archive door opens

PR #455, opened 2025-09-03
Added additional check for key object before accessing its property

arnoldblake:
Fixes bug #454 by checking for they key object before accessing its discard_after_use property.


Phazorknight (approved):
Great catch and good fix. thank  you!

Issue #456, opened 2025-09-03
Lab shooting range targets not destroyed when dart is thrown at them


arnoldblake:
**Cogito and Godot Engine Version:**
Godot 4.4.1, Nightly

**Description:**
In the lab, when throwing the yellow darts from the table at the targets, you receive this error in the debugger and the targets are not destroyed

E 0:00:36:124   cogito_projectile.gd:112 @ deal_damage(): Error calling from signal 'damage_received' to callable: 'Node(HitboxComponent.gd)::damage': Cannot convert argument 2 from Nil to Vector3.
  <C++ Source>  core/object/object.cpp:1249 @ emit_signalp()
  <Stack Trace> cogito_projectile.gd:112 @ deal_damage()
                cogito_projectile.gd:83 @ _on_body_entered()

**Reproduction steps:**
Start the game, go to the lab, pickup the yellow dart in the shooting range and throw it at a target

**Expected behavior:**
The target should be destroyed as it is when shooting them with one of the guns

arnoldblake:
In cogito_projectile.gd:24 var Direction is declared and not initialized.
Inside function deal_damage:108, Direction is assigned to bullet_direction and then the signal collider.damage_received is emitted with the parameter bullet_direction which is null causing the error.

Phazorknight:
Considered resolved with https://github.com/Phazorknight/Cogito/commit/c14503ab32a12bbbd1a847d86bd183ebb4c3c890

PR #457, opened 2025-09-03
Fixed null variable Direction assignment to deal_damage

arnoldblake:
Fixes bug #456 by removing the variable assignment of Direction which is null to bullet_direction. The darts now destroy the targets.

Tested shooting the targets with the guns and they behave as expected.


Phazorknight (approved):
Good catch!

Will clean this up a bit once it's merged but you found the fix for this.

PR #459, opened 2025-09-04
Fixed UI inventory slot selection bug

### LINKED ISSUE #216: Unrelated inventory squares are highlighted when squares are selected
The Cogito pulled from master - v4.3.beta1.mono.official [a4f2ea91a]

**Description:**
When using a mouse, selecting an object's inventory square (populated or not) causes a corresponding player's inventory slot to highlight.

I believe [this line of code](https://github.com/Phazorknight/Cogito/blob/main/COGITO/InventoryPD/UiScenes/inventory_interface.gd#L184) is the culprit. It appears it might need a check above it to see if the player is using a mouse/gamepad.

> Unrelated to the bug, I think `InputHelper` could use a `is_using_mouse()` method that does the `device_index == -1` check. That would remove the need to know the implementation details of `InputHelper`.

**Reproduction steps:**
1. Open an object's inventory
2. Click on some squares
3. See the player's inventory highlight

**Expected behavior:**
I would expect nothing to happen to any of the inventory squares except for the one I am interacting with.

**Screenshots:**

https://github.com/Phazorknight/Cogito/assets/24460210/82371e2c-44e2-4fd0-91e0-6b085b9a12ae

arnoldblake:
Fixes #216 an inventory UI bug. Corrected the project settings input map to match the control mapping in game. I have also removed the if block for checking if the event is a mouse left or mouse right as this signal path from slot_clicked is commented out.

https://github.com/Phazorknight/Cogito/blob/d7af401da776ecfca079fcafe627b6e7b1b39566/addons/cogito/InventoryPD/UiScenes/Slot.gd#L90

https://github.com/Phazorknight/Cogito/blob/d7af401da776ecfca079fcafe627b6e7b1b39566/addons/cogito/InventoryPD/cogito_inventory.gd#L37-L38

This is commented out and there are no other places in the codebase that are connected to inventory_interact.

https://github.com/Phazorknight/Cogito/blob/d7af401da776ecfca079fcafe627b6e7b1b39566/addons/cogito/InventoryPD/UiScenes/inventory_interface.gd#L179

### DIFF
--- a/addons/cogito/InventoryPD/UiScenes/Slot.gd
+++ b/addons/cogito/InventoryPD/UiScenes/Slot.gd
@@ -82,12 +82,7 @@ func _on_charge_changed():
 		charge_label.text = str(int(item_data.charge_current)) 
 
 
-func _on_gui_input(event):
-	if event is InputEventMouseButton \
-			and (event.button_index == MOUSE_BUTTON_LEFT \
-			or event.button_index == MOUSE_BUTTON_RIGHT) \
-			and event.is_pressed():
-		slot_clicked.emit(get_index(), event.button_index)
+func _on_gui_input(event: InputEvent):
 	
 	# Setting SLOT GAMPEAD INTERACTIONS HERE
 	if event.is_action_pressed("inventory_move_item"):
--- a/addons/cogito/InventoryPD/UiScenes/inventory_interface.gd
+++ b/addons/cogito/InventoryPD/UiScenes/inventory_interface.gd
@@ -157,7 +157,7 @@ func _slot_on_mouse_exit():
 
 
 func update_grabbed_slot_position():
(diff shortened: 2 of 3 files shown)


arnoldblake:
I was lazy and did not test controller input. I have fixed this issue and re-implemented the call to get_focus() with one modification that we have used the .bind() method when connecting the signal so that we can pass in the appropriate "inventory_ui" to the event handler on_inventory_button_pressed. Since this event handler is serving the dual purpose of player inventory and external inventory button presses we need to know from which ui the button was pressed.

arnoldblake:
There is a new UI bug that I noticed now that I have used the controller that when you pickup an item it seems to focus the first inventory slot. I can try and tackle this in another PR.

Issue #469, opened 2025-09-30
Initial menu UI not scaled correctly


lucaleonardi:
**Cogito and Godot Engine Version:**
- Cogito 1.1.2
- Godot 4.4.1

**Description:**
On project start, if I go to options from the menu and edit game graphics, I can't apply my changes because the "Apply button" is outside of my visible screen

**Reproduction steps:**
1. Download / clone project
2. Boot it up in Godot 4.4.1
3. Startup the project with F5
4. From main menu, go to options and change any graphics option

**Expected behavior:**
I would expect the UI to scale properly, according to my screen size, and being centered. That would avoid the button from being cut outside of visible screen.

**Screenshots:**
<img width="1919" height="1129" alt="Image" src="https://github.com/user-attachments/assets/fa3396fd-50bc-40c2-8b07-d8ca0b8caaca" />

mrezai:
@1000nettles there are still some issues at least on Linux(Ubuntu 22.04) with ultra wide monitor:

Open user data folder and remove all files for a clean run:

<img width="421" height="299" alt="Image" src="https://github.com/user-attachments/assets/967b1921-6d6c-416e-99f2-862246c62832" />

First run:

https://github.com/user-attachments/assets/23c3a30e-3297-442f-8f81-f7cfc5e34330

Second run, by resizing the window a little, UI scaled properly :

https://github.com/user-attachments/assets/e5d29810-b26e-4a84-8d60-1b59426f1ed2

Third run(UI scale is correct or at least better):

https://github.com/user-attachments/assets/12673a08-5f3a-4d90-9e3d-9a7e4aac7539

Next runs will be like second(bad UI scale) or third run(better UI scale) randomly.

PR #472, opened 2025-10-06
Various options menu / resolution fixes

1000nettles:
Hi there,

This is a pull request which should solve the following issues pertaining to the options menu and beyond:

* The big one: don't automatically set the resolution to the 0th resolution in the list of resolutions when you first play the game
  * This is fixed by not loading the values in the dropdown in the options menu if we haven't loaded the config yet
* In full screen mode (exclusive and non-exclusive), the actual "game view" is not stretched to full screen and instead lives in the top-left corner of the screen
  * This is fixed by only setting the size of the window if we're in "windowed" mode
* Full Screen and Exclusive Full Screen index positions are inconsistent
* Add borderless properly if you have selected borderless

This pull request does not:

* Set the resolution dropdown list to match the current resolution automatically
* Set the window mode to match the current window mode

To test, make sure you're not embedding the game in the editor.

Please let me know if you have any questions / comments / concerns! This should at least make it so people aren't locked out of changing their resolutions and have fullscreen weirdness. 😄 This should solve #469.

----

As an aside, I wonder if it's worth swapping out from EasyMenus to some other dependency? When I was going through this, there were a bunch of issues which I didn't even touch. Additionally, it looks like it hasn't been touched in 3 years. Just some food for thought!

### DIFF
--- a/addons/cogito/EasyMenus/Scripts/OptionsTabMenu.gd
+++ b/addons/cogito/EasyMenus/Scripts/OptionsTabMenu.gd
@@ -328,7 +328,9 @@ func load_options(skip_applying:bool = false):
 	window_mode_option_button.selected = window_mode
 	resolution_option_button.selected = resolution_index
 	
-	if !skip_applying:
+	# Only apply window mode + resolution + refresh when a config actually exists,
+	# and when we're not skipping applying.
+	if !skip_applying and have_cfg:
 		anti_aliasing_2d_option_button.emit_signal("item_selected", msaa_2d)
 		anti_aliasing_3d_option_button.emit_signal("item_selected", msaa_3d)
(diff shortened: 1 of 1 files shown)


Phazorknight:
Hey, thanks for this, will take a look.

> I wonder if it's worth swapping out from EasyMenus to some other dependency? When I was going through this, there were a bunch of issues which I didn't even touch.

EasyMenus was used as a starting framework and at this point has been modified a lot, hat's also the reason it's not listed as its own addon but as part of Cogito. Even if the original EasyMenus would update, I would probably not "pull" the updates as it would break a lot of things. 😅 
If you have specific suggestions or issues for the menus etc, feel free to open an issue or discussion. I'm definitely up to rework parts of this and have some ideas of my own on what to tackle next.

PR #478, opened 2025-10-14
Wieldable container and wieldable interaction

Phazorknight:
Currently WORK IN PROGRESS, this PR contains the following:

New interaction component:
**Wieldable interaction** - checks if the player is wielding a specific item and calls specific methods if true.

New item type:
**ContainerItemPD** - a bit niche, used for items that can contain different contents, meant to be used as Wieldables. Examples would be bottles, glasses or mugs that could hold a variety of liquids.

New wieldable:
**WieldableMug** - coffee mug that can be filled with coffee or water. Currently only coffee is supported and while a drinking animation plays, nothing more happens. WORK IN PROGRESS.

Wanted to put this PR in, so people can see the current approach/implementation of this.


Phazorknight:
Getting close to finishing this.
Still need to implement the other WieldableInteraction functions and then it should be good.

Phazorknight:
Ok, I'm marking this ready for review.

One known issue:
- The pickup mug seems to have the instance issue where each mug shares the same resource instance. Usually this gets fixed by marking it "Local To Scene" but for some reason that doesn't work on this one. I am not sure where the error is, but I probably can't see the forest for the trees. If someone could take a look / fix why the mugs share their contents, that would be fantastic.
Would love to have a water and a coffee mug to switch between.

Gromnax:
Speaking of this, I have been trying to make a remake of Space Station 13 in Godot for a while and that's the main reason for me basing my project off cogito.

In SS13, basically every click could be translated by "I am using what is currently in my active hand on what I am I clicking on". That's basically what I'm trying to reproduce in Cogito here

https://www.youtube.com/watch?v=K89zaxVvCLY
I tried to explain it in here. Dunno if it's super clear, feel free to ask if it isn't. But basically, what I'm trying to say is, wouldn't it be more simple to have, for instance, an interaction component using tool attributes as parameters, a barehand interaction just being a special case?

This is just my two cents of course

Gromnax:
I didn't expect an answer so quickly! Thank you for your message, I think you're absolutely right.

PR #481, opened 2025-10-16
Aiming system

luiscarloscontreirassilvestre-art:
The AimingSystem is designed to be attached to a Node3D that is a child of a Camera3D. It uses a RayCast3D to detect enemies within a specified range, updates a crosshair UI to reflect targeting status, and applies visual effects (like highlighting) to detected enemies.


luiscarloscontreirassilvestre-art:
Thank you, I'll take a look at this issue. But now I've added more options. The code is basically a detector for groups, but it has other capabilities.

PR #484, opened 2025-10-18
Ensure `content_scale_size` is always zero

1000nettles:
As mentioned by @mrezai in #469, in Ubuntu (and I assume other configurations and operating systems), we see that after adjusting the windowed resolution, quitting, and re-opening, the scale of the content is remaining at the default viewport size: 1920x1080. Once resized, this is fixed, but upon initial load it should respect the resolution selected for the window.

The issue stems from setting the `content_scale_size` in the logic dealing with screen resolutions. It's worth checking out the docs for [content_scale_size](https://docs.godotengine.org/en/stable/classes/class_window.html#class-window-property-content-scale-size), but essentially we were previously explicitly setting this based on the resolution selected in windowed mode. This is unnecessary because we're already handling scaling based on the selected window resolution, and the UI scale changes. Setting this to (`Vector2i(0,0)`) means we have one less scaling configuration in the mix which is nice because handling resolutions is already so complex.

We also are removing some code I tripped on which is not wired up to anything (as far as I could tell.)

**To test:**

1. Delete your user directory for the Cogito project
2. Launch the game
3. Make sure you're in Windowed mode
4. Select a resolution that's lower than your current resolution, apply changes
5. Close the game
6. Open the game again
7. Make sure the windowed resolution is correct, and the content isn't scaled to be at 1920x1080 (aka, doesn't look enormous and flow out of the window)

### DIFF
--- a/addons/cogito/EasyMenus/Scripts/OptionsTabMenu.gd
+++ b/addons/cogito/EasyMenus/Scripts/OptionsTabMenu.gd
@@ -60,14 +60,14 @@ const RESOLUTION_DICTIONARY: Dictionary = {
 	"1024x576 (16:9)": Vector2i(1024, 576),
 	"1024x640 (16:10)": Vector2i(1024, 640),
 	"1024x768 (4:3)": Vector2i(1024, 768),
-	"1152x648 (16:9)": Vector2(1152, 648),
+	"1152x648 (16:9)": Vector2i(1152, 648),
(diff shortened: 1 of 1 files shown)


mrezai:
@1000nettles thanks, I'd tested it and main menu scale is correct at the start.

There are other issues, I'll open a new discussion about them tomorrow.

PR #487, opened 2025-10-25
Starting a new game - refinements

Phazorknight:
This branch introduces the following changes/updates:
- New game scene is now set in the CogitoSettings resource
- On death, if no save is present, players can directly start a new game
- NEW RESOURCE: Cogito World State - this holds a world_dictionary. World state on new game start now also gets loaded form CogitoSettings. This enables devs to create and easily swap out different world states for testing/developing.

KNOWN ISSUES:
- When player dies with no save present, starting a new game opens a loading screen that stays stuck at 0%
- We've run into this issue before when trying to load from player death, something in the loading screen script or CogitoSceneManager doesn't get set up or gets set up in the wrong order when trying to load directly.
- Going back to the main menu and starting a new game from there seems to work.

### DIFF
--- a/addons/cogito/EasyMenus/Scripts/pause_menu_controller.gd
+++ b/addons/cogito/EasyMenus/Scripts/pause_menu_controller.gd
@@ -155,7 +155,6 @@ func _on_save_button_pressed() -> void:
 	CogitoSceneManager.save_player_state(CogitoSceneManager._current_player_node,CogitoSceneManager._active_slot)
 	#CogitoSceneManager.save_scene_state(CogitoSceneManager._current_scene_name,CogitoSceneManager._active_slot)
 	CogitoSceneManager.save_scene_state(CogitoSceneManager._current_scene_name,"temp")
-	CogitoSceneManager.delete_redundant_scene_states(CogitoSceneManager._active_slot)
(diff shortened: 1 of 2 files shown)


Phazorknight:
Oh I see.
So when I tried to replicate, the old node did get removed for me 
(as should happen in loading_screen.gd:56  `current_scene.queue_free() # Removing previous scene.` )

But the new generated node name still came to pass and with that, the wrong/bad scene save state naming.
While we might be able to fix the order this happens here to avoid the node name issue, I'm wondering if it might be better for the save system in general if we make it independent of node names. My idea would be to add a level scene name property to cogito_scene.gd and use this in the save data instead. 🤔

mrezai:
> I'm wondering if it might be better for the save system in general if we make it independent of node names. My idea would be to add a level scene name property to cogito_scene.gd and use this in the save data instead. 

It seems a good idea to me.
I think maybe it would be a good idea to use [_get_configuration_warnings](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-private-method-get-configuration-warnings) for important properties like this one.

Phazorknight:
Hey, great work! I gave it a go and didn't run into any issues (though will try to do a bit more testing).
I definitely prefer this to adding a new property to cogito scenes.

Also quickly cleaned up the signal handling on the death screen new game button.

PR #488, opened 2025-10-29
Improve loading and stuttering in demo project by turning off emission

1000nettles:
Hey! I was playing around with Cogito and something that's bothered me about the demo was the long initial load time and the scene transition time (going into the laboratory) in the demo project. I figured it was just some overhead in Godot, or the fact that I'm on a Mac. I did some digging into shader recompilation and I figured out that the issue is that every single material seems to have emission turned on. I turned it off for everything (except a few items where it makes sense to keep it on) and the loading times are almost non-existent for me.

Not sure if you wanted emission on but I figured it's easy enough to open a PR setting them to `false`. It may be worth having this change as people trying out Cogito might think it's introducing some heavy overhead when really it's just trying to load the emission. This was the conversation that lead me to dig into this on the Godot Cafe Discord:

Screenshot 2025-10-28 at 7 51 26 PM


Phazorknight:
> something that's bothered me about the demo was the long initial load time and the scene transition time (going into the laboratory) in the demo project.

Yeah, I assume this is heavily machine dependent. On my machine (Win10, Nvidia card), going from the Lobby to the Lab takes about 3 seconds in the loading screen - I appreciate insights in how it runs on other machines as that's of course hard for me to test.

> I figured it was just some overhead in Godot, or the fact that I'm on a Mac.

It might be how the Mac handles things. I've switched between the current _main_ branch and this PR and my load times have been pretty much identical  (and so have overall frame times).
I want to give this a quick try on my linux laptop out of curiosity, otherwise I have no issue merging this in after I take a closer look. There's only a handful of objects that use the emission property (lamps and buttons).

1000nettles:
@Phazorknight awesome, thanks for taking a look! Something I should have mentioned earlier is it's worth deleting the user data folder (`Project -> Open User Data Folder`) whenever testing the changes. I assume some shaders are getting cached, because the first run through the demo on `main` feels sluggish and stuttery, but when I run the demo again it's fine.

1000nettles:
Thanks for taking a look, much appreciated. I gave this another round of testing on my Mac, and for some reason, after clearing all caches and loading up `main` again, Cogito works fine without any stuttering or long load times. 🤔 It's really odd because this was a consistent issue for me previously when I was working on other PRs.

I am going to take some time this weekend to dig in a bit deeper to get to the root of the problem and make sure that removing emission is actually improving performance with the Metal renderer. If it _is_ the emission property, I will add the emission back to those items that you mentioned.

1000nettles:
My apologies for the runaround here. I've narrowed down the issue and it is definitely a Godot on Mac issue, where it pre-compiles the shaders and can take a long time on first load. The issue still persists on this branch when I blow away my shader cache (which btw, on Mac is not in the user data, but instead in `/var/folders/*/*/*/com.mycorp.*/com.apple.metal*`)

More information about this issue on Mac can be found here: https://github.com/godotengine/godot/issues/106757

Not really much we can do in Cogito for now. Sorry for wasting your time!

PR #493, opened 2025-11-25
Allow cancelling gamepad remapping via mouse and keyboard

1000nettles:
Hey! In this PR, we are addressing a bug that if you accidentally try to remap a gamepad input and have no gamepad connected, you're unable to cancel the remapping. The introduced changes now allow a user to hit their Escape key, or click anywhere to cancel the remapping.

We also adjusted the keyboard remapping functionality to match the changes in the gamepad remapping functionality.

To test, try to remap a gamepad input. Hit either Escape or click with the mouse to exit the remapping.

This was originally brought up by @mrezai in https://github.com/Phazorknight/Cogito/discussions/486.

### DIFF
--- a/addons/cogito/EasyMenus/Components/gamepad_bind_button.gd
+++ b/addons/cogito/EasyMenus/Components/gamepad_bind_button.gd
@@ -35,28 +35,44 @@ func _input(event):
 	if !is_remapping:
 		return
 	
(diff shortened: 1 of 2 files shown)


mrezai:
Thanks!
@Phazorknight Tested, it works as expected but I'm not familiar with this part of code, please review it, thanks.

PR #494, opened 2025-12-04
1.1.5 Release Candidate

Phazorknight:
Hey everyone,

I've created this PR to finalize the next release, 1.1.5.

Purpose of this release:
- Create a new base point for new/future issues and PR.
- Put this release on to the Godot Asset Store beta.
- Have it as final release on the GitHub repo before moving to Codeberg.

Included features/updates:
- Removed legacy demo scene.
- Removed Cogito Basic Enemy (as is now superceded by Cogito NPC)
- Some minor Cogito NPC improvements ( ragdoll spawn, improved pathfinding)
- Added a credits scene from the main menu
- Some minor tweaks here and there.

Note:
I've quickly tested this with Godot 4.5.1 and ran into no issues, but I'll still leave this for Godot 4.4 for now as a last release supporting **4.4**. Next release will be 4.5.1.


mrezai:
Credits issues:
- In low resolutions credits never ends.

https://github.com/user-attachments/assets/656952cf-1c2e-4fdb-8ae7-75684e911be5


- In high resolutions credits' text appears with delay for example in 3440x1440 it appears after 8 seconds.

Phazorknight:
@mrezai Thanks for taking a look!

> it seems some changes hadn't committed:

This seems odd, as I don't have an asset or dir named like that. In general also /asset/ dir is part of .gitignore. I suspect this snuck into a scene as a remnant from some other commit/user.


> In low resolutions credits never ends.

Will take a look. Both of these issues seem to come from the credits Header/Footer, which are the nodes that create the empty spaces before/after the credit text label itself. They get sized by grabbing the screen resolution, but I think this might not work correctly, leaving them at 1920x1080, which is to small for a higher res and too large for low res. In the low res case, this might cause the end condition to not be met.

Phazorknight:
@mrezai I've updated the credits, which should hopefully fix the resolution scaling issue, please take a look.

mrezai:
I've pushed some small fixes.

All my reported issues are fixed.

There is another issue related to credits, its text hides a little late and overlaps with main menu:

https://github.com/user-attachments/assets/dcc68561-ff48-4e11-b731-473b87c0734b

Another issue: Bindings tab's content automatically scrolls a little every time options menu opens.

https://github.com/user-attachments/assets/4e70b189-d30d-4778-a04f-ecd6f779c4bf

OvercastInteractive:
Tested in Godot 4.5

- I also experience what @mrezai found with the credits overlapping with the main menu when escaping (Esc key) during credits scroll
- After saving and loading, I found an issue where quest subtext doesn't appear in the quest UI:
Cogito_missing_quest_subtext

luiscarloscontreirassilvestre-art:
The problem is confirmed: the quest subtext (quest_description) does not persist after loading the save because the quest_description variable does NOT have @export in CogitoQuest.gd.

Why does this happen exactly?

Before saving:
Quest "Deliver" → complete() executed → quest_description = quest_description_completed ("We got the data...")
Saves the temporary state in memory.
UI displays correctly.

When saving:
ResourceSaver saves ONLY @export variables to the .res file.
quest_description (without @export) → IGNORED → not saved.

When loading:
Loads from .res → quest_description = "" (default value).
Adds to the completed group without calling complete() again.
UI shows empty subtext for "Deliver the CEO's laptop".

"Steal" displays "You found..." because it likely has custom logic that is dynamically set (but this could also fail in future saves).

Result in screenshots:

✅ "Steal" (active): custom description persists (for now).

❌ "Deliver" (completed): subtext disappeared.

Phazorknight:
Just merged in @luiscarloscontreirassilvestre-art 's PR (Thank you!) that should fix the quest display issues (and improve quests in general).

luiscarloscontreirassilvestre-art:
E 0:00:08:214   cogito_npc.gd:129 @ update_animations(): Type mismatch between initial and final value: Vector2 and int
  <C++ Erro>    Method/function failed. Returning: false
  <Origem C++>  scene/resources/animation.cpp:5688 @ validate_type_match()
  <Rastreamento de Pilha>cogito_npc.gd:129 @ update_animations()
                npc_state_patrol_on_path.gd:39 @ _physics_process()


	if rel_velocity_xz.length_squared() > 0:
		animation_tree.set("parameters/Movement/blend_position", rel_velocity_xz)
	else:
		animation_tree.set("parameters/Movement/blend_position", 0)

**correction**


	if rel_velocity_xz.length_squared() > 0:
		animation_tree.set("parameters/Movement/blend_position", rel_velocity_xz)
	else:
		animation_tree.set("parameters/Movement/blend_position", Vector2.ZERO)

**tested**
Godot 4.4, 4.5, 4.6-beta1

Phazorknight:
>Type mismatch between initial and final value: Vector2 and int

@luiscarloscontreirassilvestre-art  good catch! Just fixed this.

mrezai:
I've pushed fixes to improve focus on options' in OptionsMenu, mostly useful for gamepad.

There is still problem in how focus works in option and pause menus. Using gamepd focus could move to quick slots when these menus are open:

https://github.com/user-attachments/assets/5b74b4e6-8e45-4429-bd73-3e5d3692eaba

Phazorknight:
@mrezai Thanks, will take a look.

> There is still problem in how focus works in option and pause menus. Using gamepd focus could move to quick slots when these menus are open:

Yeah... these are giving me a headache at times, not just in the options. I might try manually changing their FocusMode, but not sure that'll work as I picture.

Phazorknight:
> Using gamepd focus could move to quick slots when these menus are open

@mrezai: I was able to fix this, please take a look.

@luiscarloscontreirassilvestre-art also updated the NPC ragdoll mannequin to use 6DOF on the skeleton joint constraints - looks less ridiculous now 😁

Phazorknight:
@luiscarloscontreirassilvestre-art thanks for your suggestions, but I'd like to focus on polishing/finishing this PR **as is**, not adding more features. I definitely want to work on improving the CogitoNPC in the future though in a separate update.

luiscarloscontreirassilvestre-art:
**In my opinion, the main menu needs improvement.**


Screenshot_20251217_180440

Screenshot_20251217_180503
 

Screenshot_20251217_180523

mrezai:
@Phazorknight I've tested the fix for quick slots focus issue, it works correctly, thanks.
I've pushed a small fix for inventory focus issue when it's opened for the first time.

mrezai:
> In my opinion, the main menu needs improvement.

I've been working on this issue and dynamic save/load functionality, I think it will be ready for the next release(1.2):


https://github.com/user-attachments/assets/74509f6a-ef22-4848-82ab-3eac77835411

Phazorknight:
I've cleaned up my recent changes a bit and think it's improved, but there might be some edge cases I've missed so I've decided to check my updates in.

@mrezai coincidentally also just pushed a very similar update in #505 , so there's currently some overlap between that PR and this one.

luiscarloscontreirassilvestre-art:
**NPC  Investigate State** 

The new "Investigate" state allows NPCs to intelligently and cooperatively investigate sounds/events.
Behavior

OBSERVING — The NPC stands and observes the location for observe_duration seconds
MOVING_TO_LOCATION — The NPC moves toward the investigation point at move_speed_multiplier × normal speed
INVESTIGATING — The NPC stays at the location and keeps observing until the time runs out
COMPLETE — Returns to the previous state

Exported Properties

investigate_animation (String): Animation to play during investigation (default: "alert")
observe_duration (Float): Initial observation time in seconds (default: 2.0)
investigation_distance (Float): Default distance for starting an investigation (default: 15.0)
npc_group_distance (Float): Distance within which nearby NPCs are called to join (default: 8.0)
investigation_timeout (Float): Maximum investigation duration in seconds (default: 15.0)
move_speed_multiplier (Float): Speed multiplier while moving to the location (default: 0.8)

How to Use
Trigger Investigation via Code

`# Set the location to investigate`
`npc.set_meta("investigate_location", Vector3(10, 5, 20))`
`# Switch to the investigate state`
`npc.States.goto("investigate")`


Nearby NPCs Automatically Join
When one NPC starts investigating, all NPCs within npc_group_distance are automatically notified and will also investigate the same location.

Integration with Other Systems
The state is fully compatible with:

Patrol System
Pursuit System
Combat System
Automatic state transitions

It can be triggered by security cameras, event detection systems, or any other game event trigger you implement.

mrezai (approved):
I've tested latest changes, looks good to me.
@Phazorknight I think you could merge this PR. There are other bugs but if we want to fix them all, this version never will  be released! We can fix them in next version.

PR #505, opened 2025-12-28
Interactions fixes and improvements

### LINKED ISSUE #503: Wieldable actions aren't interrupted when detecting interaction objects using the same input action
Cogito v1.1
Godot v4.4.1 stable

**Description:**
See title. The wieldable isn't told to ignore an input when attempting to interact with an object in the world that uses the same input action, causing two separate, and potentially unintended actions, to occur simultaneously.

**Reproduction steps:**
1. Equip the foam pistol and fire it to reduce your current ammo/magazine count below the max amount.
2. Use the ammo vending machine and pick up some foam bullets, but don't reload.
3. Use the ammo vending machine again (either button) to see both a reload action and vending machine action triggered irrespective of each other.

**Expected behavior:**
Expecting wieldable actions that share the same input action with a detected interaction node should be ignored or interrupted.

**Video:**
https://github.com/user-attachments/assets/6b1995e7-aab4-424f-935a-e49285ad3ed8

mrezai:
I don't know, maybe there is a better solution out there, so please review cautiously!
Fixes #503 and https://github.com/Phazorknight/Cogito/pull/494#issuecomment-3679895001.
- Add AllowedInteractionType to control interactions when inventory is open, set it to ALWAYS for fridges.
- Fix mouse actions for wieldables.

### DIFF
--- a/addons/cogito/Components/Interactions/BasicInteraction.gd
+++ b/addons/cogito/Components/Interactions/BasicInteraction.gd
@@ -1,6 +1,9 @@
 extends InteractionComponent
+class_name BasicInteractionComponent
 
 signal basic_signal
+
+@export var allowed_interaction_type : AllowedInteractionType = AllowedInteractionType.INVENTORY_IS_CLOSED
 
 @onready var parent_node = get_parent() #Grabbing reference to parent
 
--- a/addons/cogito/Components/Interactions/InteractionComponent.gd
+++ b/addons/cogito/Components/Interactions/InteractionComponent.gd
@@ -5,6 +5,11 @@ class_name InteractionComponent
 
 signal was_interacted_with(interaction_text,input_map_action)
 
+enum AllowedInteractionType {
+	ALWAYS,
+	INVENTORY_IS_CLOSED,
+}
+
 ## The input map action string as defined in the Godot Project settings input map.
 @export var input_map_action : String
 ## The text that gets displayed in the HUD interaction prompt.
--- a/addons/cogito/Components/PlayerInteractionComponent.gd
+++ b/addons/cogito/Components/PlayerInteractionComponent.gd
@@ -62,6 +62,7 @@ var is_wielding: bool:
 	get: return equipped_wieldable_item != null
 	
 var player_rid
+var _player_hud: CogitoPlayerHudManager
 # Briefly interrupt quickslot cycling after an interactable is unseen
 var can_cycle_quickslots: bool = true
 @onready var cycle_quickslots_interrupt_timer: Timer = $Timer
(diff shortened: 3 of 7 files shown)


Phazorknight:
@mrezai thank you so much for this.

Co-incidentally (or unsurprisingly) this is very similar to what I had worked on (and just checked in) on #494 . As we've been working on this at the same time, we might cross some wires with this PR and #494.

- AllowedInteractionType: I hope we're able to get input/interactions to behave as we want without having to add another property to them, and I'd personally like to avoid this as it increases complexity and allows for more "user-errors" by potentially letting devs set this wrong.
- To be honest I'm also not 100% sure I understand how this is used in the example, so please correct me: Is this to let the player close the cogito_containers / fridges on gamepad? As of without this, I think the fridges can be opened with _interact_ but for closing, the player needs to press the _inventory_ key.


Slot.gd / Quickslot updates:
Nice one cleaning these up.

mrezai:
> To be honest I'm also not 100% sure I understand how this is used in the example, so please correct me: Is this to let the player close the cogito_containers / fridges on gamepad? As of without this, I think the fridges can be opened with interact but for closing, the player needs to press the inventory key.

You're right, AllowedInteractionType added to let the player close cogito_containers with the _same_ interaction key and with other changes to impose mouse/keyboard consistent interactions.
Test like this:
- Press F to open the fridge, without moving the mouse, press F again, fridge will be closed.
- Press F to open the fridge, move the mouse a little(move pointer over one of the slots' item), press F again, fridge will not be closed.
- Press F to open the fridge, move the mouse outside of the container's UI, press F again, fridge will be closed.

There are two solutions:
- Using current implementation for more natural/convenient open/close interactions.
- Removing AllowedInteractionType and making mouse/keyboard interaction to behave like gamepad(open by interaction key, close by inventory key). This has another issue, fridge prompt UI still shows "Close" when container's UI closed by pressing the inventory key.

f1

Phazorknight:
Sorry for the late reply, and thanks for the explanation!

I think the AllowedInteraction type in a way has some overlap with the "Ignore Open GUI" bool of _InteractionComponent_, which was made to ignore the interaction input event if a GUI is open.

In general, I've been thinking a bit over these niche cases we encounter (allowing interactions to be triggered or not while another interaction openes a window or whatnot) a bit and looking at how other games handle this. 
What I've encountered is a singular "Close" button/key for containers or interfaces that "open", and making sure that that's an intuitive one. In most games that's gamepad B / keyboard ESC.

I think this would be doable for Cogito with only some smaller changes, but this would also NOT allow players to open/close a container with the same input map action anymore.
Would be something to consider.

mrezai:
> What I've encountered is a singular "Close" button/key for containers or interfaces that "open", and making sure that that's an intuitive one. In most games that's gamepad B / keyboard ESC.

It would be a good idea that we use familiar key bindings.
I'll close this PR because its main part currently merged in #494. We can improve interactions in next release.

PR #507, opened 2026-01-01
Add patrol path check before setting target position

luiscarloscontreirassilvestre-art:
0:0:8.705 | Invalid assignment of property or key 'target_position' with value of type 'Nil' on a base object of type 'NavigationAgent3D'.


Phazorknight (approved):
Thanks for catching and fixing this!

PR #511, opened 2026-01-12
auto_unequip_and_reequip_for_carry_drop_throw

OvercastInteractive:
- Automatically unequips the current wieldable when attempting to carry an object
- When leaving a carried object (drop, throw or leave) reequips the current wieldable
- Creates a more seamless & effortless transition between wielding and carrying. Similar in functionality & behavior to Half-Life 2.


Phazorknight:
Hey, thanks for this - actually had implementing somethign like this on my to-do list so I'm very happy to see this.
Would you mind putting this PR on the new 1.1.6-rc branch here? https://codeberg.org/Phazorknight/Cogito/src/branch/1.1.6-rc

OvercastInteractive:
Alright, I set up my codeberg account and will try to work on this branch soon for 1.1.6-rc.

I discovered a bug where spamming carry and drop/throw input can allow the player to carry an object while wielding, which likely has something to do with animation playback for unequip or equip and some of the await timeout calls we use in wieldable handling functions in PIC, so I'll aim to fix those before publishing the PR.

